PR #2, opened 2018-11-30
Remove `jackdaw.topic` namespace from serdes resolution config

99-not-out:
Removing some left over use of the `jackdaw.topic` namespace in the serdes resolution code. Topic config for serdes resolution should look like:

```
{...
 :key-serde ...
 :value-serde ...
...}
```

Added a test namespace for the top level serdes interface.

### DIFF
--- a/src/jackdaw/serdes.clj
+++ b/src/jackdaw/serdes.clj
@@ -56,18 +56,19 @@
                             {:topic topic-or-kw})))))
 
   ([mapping topic-or-kw default-ctor]
-   (let [key (or (::type topic-or-kw)
-                 topic-or-kw)]
-     (when-not (keyword? key)
+   (let [k (or (::type topic-or-kw)
+               topic-or-kw)]
+     (when-not (keyword? k)
        (throw (ex-info "Failed to find a legal mapping `key` for `topic-or-kw`"
                        {:topic-or-kw topic-or-kw
-                        :key key})))
-     (-> (get mapping key default-ctor)
+                        :key k})))
+     (-> (get mapping k default-ctor)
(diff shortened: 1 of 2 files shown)


src/jackdaw/serdes.clj
```diff
@@ -56,18 +56,19 @@
                             {:topic topic-or-kw})))))
 
   ([mapping topic-or-kw default-ctor]
-   (let [key (or (::type topic-or-kw)
-                 topic-or-kw)]
-     (when-not (keyword? key)
+   (let [k (or (::type topic-or-kw)
```
99-not-out: renaming `key` to `k` to stop it shadowing `clojure.core/key`

src/jackdaw/serdes.clj
```diff
@@ -110,11 +109,11 @@
             ;;   parameterizing the ctor on the thus altered topic config is
             ;;   fairly janky. There's probably a better way to generally
             ;;   parameterize serdes, but leaving this as is for now.
-            ::key-serde
+            :key-serde
```
99-not-out: Do we really want to overwrite the serde key with the resolved implementation? The name spacing prevented that in nascent implementation - `:jackdaw.topic/value-serde` in and `:jackdaw.serdes/value-serde` out

99-not-out:
Ok, good to know the direction we want to go in.
We need a home for the serdes code though (or more specifically, a home for the code which deals with the confluence of topic config / schemas / kafka interop, which is a shared concern for our use case at least)!

99-not-out:
Closing based on moving the serdes glue out somewhere else (please lets decide where)

PR #4, opened 2018-12-03
correct kstream->ktable left-join fn args

mbragg02:
Corrects the kstream->ktable left-join fn arity which accepts the serdes, as the [contract](https://kafka.apache.org/20/javadoc/org/apache/kafka/streams/kstream/KStream.html#leftJoin-org.apache.kafka.streams.kstream.KTable-org.apache.kafka.streams.kstream.ValueJoiner-org.apache.kafka.streams.kstream.Joined-) now requires a `Joined` instance

### DIFF
--- a/src/jackdaw/streams.clj
+++ b/src/jackdaw/streams.clj
@@ -57,8 +57,8 @@
   key."
   ([kstream ktable value-joiner-fn]
    (p/left-join kstream ktable value-joiner-fn))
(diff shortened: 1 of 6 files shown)


arrdem:
We're still figuring our external release cadence out, but we should be able to get this deployed later today.

mbragg02:
Thanks @arrdem / @kidpollo / @99-not-out

PR #15, opened 2018-12-17
Upgrade to Kafka 210 and CP 510

creese:


test/jackdaw/serdes/avro_test.clj
```diff
@@ -483,5 +483,5 @@
     (is (= (String. (round-trip serde "bananas" (.getBytes "hello")))
            "hello"))
 
-    (is (thrown? org.apache.kafka.common.errors.SerializationException
```
cddr: Do you think the fact that this has changed means we have no business including a test for it? Can you remember the original motivation?

creese: I really don't.

PR #16, opened 2018-12-18
[wip] add cloverage and cloverage step

AndreaCrotti:


.circleci/config.yml
```diff
@@ -24,20 +23,61 @@ jobs:
...
+            - libs-cache-{{ checksum "project.clj" }}-{{ checksum ".circleci/config.yml" }}
+            - libs-cache-{{ checksum "project.clj" }}
+            - libs-cache-
+
+      - run: nohup confluent start
+      - run: confluent status
+      - run:
+          name: run code coverage
+          # need also `shared` because of the repositories setting
+          command: lein with-profile test,shared cloverage --codecov
```
AndreaCrotti: a bit annoying to do this but it doesn't work otherwise, since specifying the profile doesn't make it pick up the shared profile

AndreaCrotti:
This generates some interesting errors like:

```Loaded  jackdaw.client.partitioning  .
Reflection warning, jackdaw/test.clj:1:146 - reference to field close can't be resolved.
Reflection warning, jackdaw/test.clj:1:146 - reference to field close can't be resolved.
java.lang.ClassCastException: clojure.lang.Keyword cannot be cast to clojure.lang.Compiler$LocalBinding```

which are caused by the `cloverage` instrumentation, and probably some vaguely weird code that breaks it.

.circleci/config.yml
cddr: Seems like you can do `lein with-profile +test cloverage --codecov`.

AndreaCrotti:
This was moved in another PR so we can close it

PR #17, opened 2018-12-19
seq->log

AndreaCrotti:
seq only takes one argument so it can't be the desired function to use
there, it looks like `log` should be the right function to use.

### DIFF
--- a/src/jackdaw/client/log.clj
+++ b/src/jackdaw/client/log.clj
@@ -12,7 +12,7 @@
 
   If fuse-fn was provided, stops after fuse-fn returns false."
   ([^Consumer consumer polling-interval-ms]
-   (seq consumer polling-interval-ms (constantly true)))
+   (log consumer polling-interval-ms (constantly true)))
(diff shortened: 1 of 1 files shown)


AndreaCrotti:
We have to merge https://github.com/FundingCircle/jackdaw/pull/18
first to get the tests running though.
Maybe a test for these functions would be good as well since I think they are never exercised atm (otherwise the bug would have been evident)

AndreaCrotti:
Yes ok @arrdem but I can't ship it if the circleci build doesn't work, we have to merge https://github.com/FundingCircle/jackdaw/pull/18 first

AndreaCrotti:
The error seems to be really just a timeout.
Currently in master this code is actually exercised

`(seq consumer polling-interval-ms seq)`

and it should actually throw an exception but it looks like it's caught somewhere.
The fact that now it actually does something instead of failing silently makes it timeout, at least that's the only explanation I could come up for now.

PR #18, opened 2018-12-19
use public lein and install needed packages for confluent

AndreaCrotti:
This will make sure that also PRs from forks are able to run tests.
The alternatives could be:

- use docker images directly instead of the confluent package
- publish our private image somewhere

(both will potentially be faster than this solution since they enable more caching, but this solution is probably the easiest and unblocks anyone who would want to contribute from a forked repo right now)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ version: 2
 jobs:
   build:
     docker:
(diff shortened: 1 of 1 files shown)


arrdem:
Need @creese or someone to merge because apparently rewriting a third of the repo doesn't qualify one to merge code.

PR #23, opened 2019-01-02
Introduce `suppress` to streams.

truemped:
Version 2.1 introduced a new operator: [suppress](https://kafka.apache.org/21/documentation/streams/developer-guide/dsl-api.html#window-final-results). It allows to only
forward the final result of a windowed aggregation.

This PR adds the `suppress` function to the streams namespace and
allows for the new operator to be used in Kafka Streams apps. It works but may not be in a final stage. Asking for feedback at this point whether this is what you have in mind or I should change something.

Best
Daniel

### DIFF
--- a/src/jackdaw/streams/interop.clj
+++ b/src/jackdaw/streams/interop.clj
@@ -10,6 +10,8 @@
             Pattern]
            [org.apache.kafka.common.serialization
             Serde]
+           [java.time
+            Duration]
            [org.apache.kafka.streams
             KafkaStreams]
            [org.apache.kafka.streams
@@ -19,8 +21,8 @@
             JoinWindows KGroupedStream KGroupedTable KStream KTable
             KeyValueMapper Materialized Merger Predicate Printed Produced
             Reducer Serialized SessionWindowedKStream SessionWindows
-            TimeWindowedKStream ValueJoiner ValueMapper
-            ValueMapperWithKey ValueTransformerSupplier Windows]
+            Suppressed Suppressed$BufferConfig TimeWindowedKStream ValueJoiner
+            ValueMapper ValueMapperWithKey ValueTransformerSupplier Windows]
            [org.apache.kafka.streams.processor
             StreamPartitioner]))
 
@@ -41,6 +43,21 @@
   (cond-> (Materialized/as ^String topic-name)
     key-serde (.withKeySerde key-serde)
     value-serde (.withValueSerde value-serde)))
+
+(defn suppress-config->suppressed
+  [{:keys [max-records max-bytes until-time-limit-ms]}]
+  (let [config (cond
+                 (not (nil? max-records))
+                 (Suppressed$BufferConfig/maxRecords max-records)
+
+                 (not (nil? max-bytes))
+                 (Suppressed$BufferConfig/maxBytes max-bytes)
+
+                 :else (Suppressed$BufferConfig/unbounded))]
+    (if-some [time-limit until-time-limit-ms]
+      (Suppressed/untilTimeLimit (Duration/ofMillis time-limit) config)
+      (-> (.shutDownWhenFull config)
+          Suppressed/untilWindowCloses))))
 
 (declare clj-kstream clj-ktable clj-kgroupedtable clj-kgroupedstream
          clj-global-ktable clj-session-windowed-kstream
@@ -395,6 +412,11 @@
      (.outerJoin ^KTable ktable
                  ^KTable (ktable* other-ktable)
                  ^ValueJoiner (value-joiner value-joiner-fn))))
+
+  (suppress
+    [_ suppress-config]
+    (clj-ktable
+       (.suppress ^KTable ktable (suppress-config->suppressed suppress-config))))
 
   (to-kstream
     [_]
--- a/src/jackdaw/streams/protocols.clj
+++ b/src/jackdaw/streams/protocols.clj
@@ -192,6 +192,14 @@
     [ktable key-value-mapper-fn]
     "Converts a KTable to a KStream.")
 
+  (suppress
+    [ktable {:keys [max-records max-bytes until-time-limit-ms]}]
+    "Suppress some updates from this changelog stream.
+    You can either specify `max-records` or `max-bytes`. If an empty map is
+    passed, the suppress will be unbounded. If `until-time-limit-ms` is set,
+    this will override the `TimeWindow` interval. Note that when relying on the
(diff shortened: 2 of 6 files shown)


99-not-out:
Hey - thanks for this Daniel.
We will do a review soon and update here.

Regards
Matt

src/jackdaw/streams/interop.clj
```diff
@@ -389,6 +400,15 @@
                  (ktable* other-ktable)
                  (value-joiner value-joiner-fn))))
 
+  (suppress
+    [_ {:keys [max-records max-bytes until-time-limit-ms]
+        :as suppress}]
+    (let [cfg (suppress-config suppress)]
+      (println cfg)
```
cddr: Shall we lose the `println` which I presume was there for debugging purposes.

src/jackdaw/test.clj
```diff
@@ -67,7 +67,7 @@
 (defn consume-all-outputs!
```
cddr: We replaced this test namespace with [the test machine](https://github.com/FundingCircle/jackdaw/pull/28) (apologies for the breaking change). There will be a conflict on this file when you update the master branch. You can resolve by simply accepting the upstream branch.

test/jackdaw/streams_test.clj
```diff
@@ -603,7 +604,45 @@
             (is (= 3 (count keyvals)))
             (is (= [1 1] (first keyvals)))
             (is (= [1 3] (second keyvals)))
-            (is (= [1 6] (nth keyvals 2)))))))))
+            (is (= [1 6] (nth keyvals 2))))))))
+
+  (testing "suppress records"
```
cddr: Shall we add another example that covers the use of a suppress config (e.g. max-records, until-time-limit etc).

src/jackdaw/streams/interop.clj
truemped: ups...

src/jackdaw/test.clj
truemped: +1

test/jackdaw/streams_test.clj
truemped: Will do. I'll try to get to it by next Monday!

truemped:
Hey, @cddr. It's been a while. I've updated the PR and added a lot more test cases. Let me know what you think. Would be great to get a new version with including this out, once it's in a state that works for  you.

Best
Daniel

.lein-failures
```diff
@@ -0,0 +1 @@
+{"jackdaw.streams-test" ["KTable"]}
```
cddr: Suspect this was not intended to be checked in?

src/jackdaw/streams/interop.clj
```diff
@@ -396,6 +408,14 @@
                  ^KTable (ktable* other-ktable)
                  ^ValueJoiner (value-joiner value-joiner-fn))))
 
+  (suppress
+    [_ {:keys [max-records max-bytes until-time-limit-ms]
```
cddr: Seems like you no longer need to destructure the `suppress` argument here.

.lein-failures
truemped: No, removed

src/jackdaw/streams/interop.clj
truemped: Yes, updated the code and also removed the `let` binding. No need for it

truemped:
Cool, thanks @cddr. Anything I still should do/help do to get it merged? Some circleci job fails, but that does not seem to be test related.

src/jackdaw/streams/protocols.clj
```diff
@@ -192,6 +192,14 @@
...
     "Converts a KTable to a KStream.")
 
+  (suppress
+    [ktable {:keys [max-records max-bytes until-time-limit-ms]}]
+    "Suppress some updates from this changelog stream. You can either specify
+    `max-records` or `max-bytes`. If an empty map is passed, the suppress will
+    be unbounded. If `until-time-limit-ms` is set, this will override the
+    `TimeWindow` interval. Note that when relying on the configured `TimeWindow`
+    the default `grace` period is `24h - window-size`.")
+
```
vijumathew: could be helpful to format the docstring after the styleguide. just ending the first line after the first sentence and starting "You can either specify..." on a newline

https://github.com/bbatsov/clojure-style-guide#docstring-summary

src/jackdaw/streams/interop.clj
```diff
@@ -396,6 +408,11 @@
                  ^KTable (ktable* other-ktable)
                  ^ValueJoiner (value-joiner value-joiner-fn))))
 
+  (suppress
+    [_ suppress]
+    (clj-ktable
+       (.suppress ktable (suppress-config suppress))))
```
vijumathew: it looks like the spec defined before was named `suppress-config` but the fn that uses it is called `suppress-config` and the config is named `suppress`. this seems a little confusing to me. 

might be helpful to name the config arg `suppress-config` to match the spec and the fn `create-suppress-config` or something. the names i suggested could definitely be better though

test/jackdaw/streams_test.clj
```diff
@@ -643,7 +644,177 @@
...
+            (is (= [1 7] (first keyvals)))
+            (is (= [1 3] (second keyvals))))))))
+
+  (testing "suppress records with a bounded buffer"
+    (let [topic-a (mock/topic "topic-a")
+          topic-b (mock/topic "topic-b")
+          topic-c (mock/topic "topic-c")
+
+          ;; if we don't set the `grace` period of the `TimeWindows`, the
+          ;; default is used: 24h - window
```
vijumathew: including this comment once seems sufficient to me. could get rid of this

src/jackdaw/streams/protocols.clj
truemped: Updated

src/jackdaw/streams/interop.clj
truemped: I've tried to make more sense of the names. The argument is now `suppress-config` and the function is `suppress-config->suppressed` which is what it's doing actually: turning a `suppress-config` into a `org.apache.kafka.streams.kstream.Suppressed`.

test/jackdaw/streams_test.clj
truemped: removed

truemped:
Hey @vijumathew, could you have another look at this?

PR #25, opened 2019-01-04
enable eastwood

AndreaCrotti:


AndreaCrotti:
There are currently several exceptions being thrown trying to run eastwood, for example:

```Exception in thread "main" java.lang.IllegalAccessError: IGlobalKTable does not exist, compiling:(/tmp/form-init2803594268547680069.clj:1:73)```

See https://circleci.com/gh/FundingCircle/jackdaw/59?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

This could be still merged but there are probably various things to fix in the code before we can get a meaningful result from eastwood

PR #28, opened 2019-01-07
Add Test Machine to Jackdaw

99-not-out:
Moving the work from @cddr into Jackdaw (all credit to @cddr).

Adds machinery to Jackdaw for running black box tests against a kafka application, either directly against the brokers or via the confluent Rest Proxy API. Allows generation of a test which can then (without modification) be run locally, in CI or against a remote environment.

Documentation will be added soon!


src/jackdaw/test/serde.clj
```diff
@@ -0,0 +1,168 @@
...
+   [jackdaw.serdes.json :as json-serde])
+  (:import
+   (org.apache.kafka.clients.consumer ConsumerRecord)
+   (org.apache.kafka.common.serialization Deserializer Serdes Serializer
+                                          ByteArraySerializer
+                                          ByteArrayDeserializer)
+   (org.apache.kafka.common.errors SerializationException)))
+
+
+;; Access to various serdes
```
99-not-out: Need to find a better home for the below section maybe (resolving serdes for topics)

src/jackdaw/test/transports/kafka.clj
```diff
@@ -0,0 +1,201 @@
...
+(defn subscription
+  "Subscribes to `topic-collection` and seeks to the end of all partitions. This
+   is usually what you want in a testing context. It's best for the test you're
+   trying to run now to ignore all the garbage created by previous tests."
+  [kafka-config topic-collection]
+  (-> (kafka/consumer kafka-config byte-array-serde)
+      (subscribe topic-collection)
+      (seek-to-end)))
+
+(defn mk-consumer-record
```
99-not-out: There is a jackdaw version of this but for a reason I forget right now I couldn't use it.

src/jackdaw/test/transports/kafka.clj
```diff
@@ -0,0 +1,201 @@
...
+     :key (.key consumer-record)
+     :offset (.offset consumer-record)
+     :partition (.partition consumer-record)
+     :serializedKeySize (.serializedKeySize consumer-record)
+     :serializedValueSize (.serializedValueSize consumer-record)
+     :timestamp (.timestamp consumer-record)
+     :topic (.topic consumer-record)
+     :value (.value consumer-record)}))
+
+(defn ^ProducerRecord mk-producer-record
```
99-not-out: There is a jackdaw version of this but for a reason I forget right now I couldn't use it.

src/jackdaw/test/transports/kafka.clj
```diff
@@ -0,0 +1,201 @@
...
+(defn close-consumer
+  [consumer]
+  (reset! (:continue? consumer) false)
+  (async/<!! (:process consumer)))
+
+(defn build-record
+  "Builds a Kafka Producer and assoc it onto the message map"
+  [m]
+  (let [rec (mk-producer-record (:topic m)
+                                (:partition m (int 0)) ;; TODO proper partitioning ...
```
99-not-out: TODO FIXME!
Currently no messages have the partition set and there is no code to try and work the partition out. Need some way to inject the correct / default partitioner into test machine writes

test/jackdaw/test/commands/write_test.clj
```diff
@@ -0,0 +1,94 @@
...
+   [jackdaw.test.serde :as serde]
+   [jackdaw.test :refer [test-machine]]
+   [clojure.test :refer :all]
+   [jackdaw.serdes.avro.schema-registry :as reg]))
+
+(def schema-registry-config
+  {:avro.schema-registry/client (reg/mock-client)
+   :avro.schema-registry/url    "localhost:8081"})
+
+(def resolver (serde/local-serdes-resolver schema-registry-config))
```
99-not-out: The above 5 lines are repeated in a few places so we should perhaps try to factor it out somewhere better.

src/jackdaw/test/transports/kafka.clj
cddr: It might be a good point to re-think the syntax of write commands a bit. Currently a write command is like `[:write topic-map k v timestamp timeout]`, which worked ok for our original specific use-case but I think we can improve on this to easily support a wider variety of use-cases.

e.g. in the examples below...

 record-map could include values for keys `:timestamp`, `:key`, `:value`, `:partition` etc
 write-option-map could be used to configure things like timeout, message based key/partition generation, or anything else we might want to configure in the future.

```
[:write topic-map record-map timeout]
[:write topic-map record-map write-option-map]
```

src/jackdaw/test/transports/rest_proxy.clj
```diff
@@ -0,0 +1,318 @@
...
+  "Builds a Kafka Producer and assoc it onto the message map"
+  [m]
+  ;; The easiest way to make this as real as possible is to build
+  ;; a real ProducerRecord and then get extract the data from it. This way
+  ;; we don't have to calculate the partition oureslves.
+  ;;
+  ;; On the other hand, this means we have a bit of an unwanted dependency
+  ;; (in this ns) on the Kafka Client API. Perhaps later it might be
+  ;; worth calculating this partition ourselves based on the supplied
+  ;; topic-metadata but there are a few wrinkles to iron out with that
```
cddr: The wrinkles mentioned here relate to the cleanup of the write command suggested [above]( https://github.com/FundingCircle/jackdaw/pull/28/files#r246106743). We want a data representation of a write that both transports can interpret.

src/jackdaw/test/transports/kafka.clj
99-not-out: Agreed - when giving it a go today a few of these came up immediately (more control over partitions / keys etc). Lets try to catch up tmrw.

src/jackdaw/test/serde.clj
99-not-out: Simplified it a bit now (removed all the Avro lookup support as it as wasn't used), this is ok to leave in I think - or maybe as its only used in the tests for the test machine, it can move under the /test folder

src/jackdaw/test/commands/write.clj
```diff
@@ -0,0 +1,68 @@
...
+  ;;  - If the topic map contains a `:key-fn`, use that function to extract
+  ;;    the key from the message
+  ;;  - If the topic map contains a `:partition-fn`, use that function to
+  ;;    determine the partition to write to (should be an fn of airity 2
+  ;;    as per `default-partition-fn`
+  ;;  - The `:key-fn` and `:partition-fn` can also be passed separately in
+  ;;    the options map `opts`.
+  ;;  - Further, the `opts` map can contain an explciit `:key` and/or
+  ;;    `:partition`, which if set will provide the values to use
+  ;; If both specified, `opts` values will override values in the topic map.
```
creese: This should probably be a doc-string.

src/jackdaw/test/transports/kafka.clj
```diff
@@ -0,0 +1,201 @@
...
+   org.apache.kafka.clients.consumer.ConsumerRecord
+   org.apache.kafka.clients.producer.ProducerRecord))
+
+(defn subscribe
+  "Subscribes to specified topics
+
+   `consumer` should be a kafka consumer
+   `topic-config` should be a sequence of topic-metadata maps"
+  [consumer topic-config]
+  (kafka/subscribe consumer topic-config))
```
creese: Is `subscribe` meant to be part of a public API? Why not just use `jackdaw.client/subscribe`?

src/jackdaw/test/transports/kafka.clj
```diff
@@ -0,0 +1,201 @@
...
+(defn seek-to-end
+  "Seeks to the end of all the partitions assigned to the given consumer
+   and returns the updated consumer"
+  [consumer & topic-partitions]
+  (let [assigned-partitions (or topic-partitions (load-assignments consumer))]
+    (.seekToEnd consumer assigned-partitions)
+    (doseq [assigned-partition assigned-partitions]
+      ;; This forces the seek to happen now
+      (.position consumer assigned-partition))
+    consumer))
```
creese: This looks similar to `jackdaw.client/seek-to-end-eager`. Can we use that instead?

https://github.com/FundingCircle/jackdaw/blob/master/src/jackdaw/client.clj#L218

src/jackdaw/test/transports/kafka.clj
```diff
@@ -0,0 +1,201 @@
...
+        (async/put! messages {:error e})))))
+
+(defn subscription
+  "Subscribes to `topic-collection` and seeks to the end of all partitions. This
+   is usually what you want in a testing context. It's best for the test you're
+   trying to run now to ignore all the garbage created by previous tests."
+  [kafka-config topic-collection]
+  (-> (kafka/consumer kafka-config byte-array-serde)
+      (subscribe topic-collection)
+      (seek-to-end)))
```
creese: This is similar to `jackdaw.client/subscribed-consumer`:
https://github.com/FundingCircle/jackdaw/blob/master/src/jackdaw/client.clj#L141

src/jackdaw/test/transports/kafka.clj
creese: Let's add a TODO so we don't forget to refactor towards that.

src/jackdaw/test/transports/kafka.clj
creese: Let's add a TODO so we don't forget to refactor towards that.

test/jackdaw/test_test.clj
```diff
@@ -0,0 +1,156 @@
+(ns jackdaw.test-test
+  (:require
+   [clojure.test :refer :all]
+   [jackdaw.serdes.avro.schema-registry :as reg]
+   [jackdaw.streams :as k]
+   [jackdaw.test :as jd.test]
```
creese: Elsewhere, we've used the equivalent of `[jackdaw.test :as jt]`. I've also seen '[jackdaw.test :as j.test]`. We should standardize on one of these three.

project.clj
```diff
@@ -1,7 +1,9 @@
 (defproject fundingcircle/jackdaw "_"
   :description "A Clojure library for the Apache Kafka distributed streaming platform."
 
-  :dependencies [[clj-time "0.13.0"]
+  :dependencies [[aleph "0.4.6"]
+                 [clj-time "0.13.0"]
+                 [org.clojure/core.async "0.4.474"]
```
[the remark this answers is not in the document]
cddr: Yep. I think that makes sense. I ended up introducing aleph to support asynchronous http requests initiated by the rest-proxy transport.

But the `go-loop` in the test-machine should be relatively portable to aleph's `d/loop` and it shouldn't change the external contract to do so.

Issue #30, opened 2019-01-07
jackdaw.client.send! delays the entire operation, rather than doing it asynchronously


99-not-out:
The intent of `send!` is to write the data but not block for the result. It uses a delay around the actual result to delay the `datafy` of the response when it arrives.

Unfortunately, the use of `delay` also wraps the `send.` call to kafka, so calling `send!` does nothing until you `deref` it, at which point you are blocked until the write succeeds. The use of the threading macro obfuscates things here also (for me at least ...)

https://github.com/FundingCircle/jackdaw/blob/a0987cb18c689c251a9b1ecd4557d5743c13d3aa/src/jackdaw/client.clj#L61

Can be fixed by moving the `send.` call outside of the `delay`, so the kafka write is started and we just delaying the datafying of the eventual result, without the threading macro use so its more explicit. i.e.

```
(let [send-future (.send ^Producer producer ^ProducerRecord record)] 
     (delay (jd/datafy @send-future)))
```

I'll raise a PR tmrw,

PR #31, opened 2019-01-08
Fix kafka send! so the send actually starts when called

99-not-out:
Fixes issue #30

### DIFF
--- a/src/jackdaw/client.clj
+++ b/src/jackdaw/client.clj
@@ -57,13 +57,13 @@
   either [RecordMetdata, nil] or [nil, Exception] respectively if the
   record was sent or if an exception was encountered."
   ([producer record]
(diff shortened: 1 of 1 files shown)


src/jackdaw/client.clj
```diff
@@ -57,13 +57,13 @@
   either [RecordMetdata, nil] or [nil, Exception] respectively if the
   record was sent or if an exception was encountered."
   ([producer record]
-   (-> (.send ^Producer producer ^ProducerRecord record)
-       deref jd/datafy delay))
+   (let [send-future (.send ^Producer producer ^ProducerRecord record)]
+     (delay (jd/datafy @send-future))))
```
99-not-out: Move the `send!` call outside of the `delay` macro, so that the send call is actually made here and not delayed until the return value of this function is deref'd

PR #42, opened 2019-01-30
add a docker compose file

AndreaCrotti:


cddr:
👍on including this but I think as you mention it should include the rest-proxy too as that is necessary to run the tests. Might also be cool to include some docs somewhere on running the tests in docker-compose.

AndreaCrotti:
Yes it's still WIP and I haven't tested it at all, just copied some config from another project so far

AndreaCrotti:
Closing this since @cddr took over it in another PR

Issue #24, opened 2019-01-03
key not used in jackdaw.client/produce!


iku000888:
produce! accepts `key` as an argument and does nothing with it.

```
(defn produce!
 "Helper wrapping `#'send!`.

 Builds and sends a `ProducerRecord` so you don't have to. Returns
 a future which will produce datafied record metadata when forced."
 ([producer topic value]
  (send! producer
         (jd/->ProducerRecord topic value)))
 ([producer topic key value]
  (send! producer
         (jd/->ProducerRecord topic value)))
 ([producer topic partition key value]
  (send! producer
         (jd/->ProducerRecord topic partition topic value)))
 ([producer topic partition timestamp key value]
  (send! producer
         (jd/->ProducerRecord topic partition timestamp topic value)))
 ([producer topic partition timestamp key value headers]
  (send! producer
         (jd/->ProducerRecord topic partition timestamp topic value headers))))
```

PR #45, opened 2019-01-31
fix produce! to accept key

sevdafc:
There seems to be a bug in the `produce!` method, it either discards the `key` argument completely or it replaces it with `topic`.


cddr:
Thanks for the patch! I think this fixes #24.

iku000888:
Indeed it does.

PR #46, opened 2019-02-01
add circleci badge

AndreaCrotti:


README.md
```diff
@@ -1,3 +1,5 @@
+[![CircleCI](https://circleci.com/gh/FundingCircle/jackdaw/tree/master.svg?style=svg)](https://circleci.com/gh/FundingCircle/jackdaw/tree/master)
```
AndreaCrotti: not sure if there is a better place to move this, and it always points only to master, but I think it's quite common to have it on top.

PR #51, opened 2019-02-05
add user-events-example

astelvida:
I thought I'd add a simple example of using a K-table. Inspired by https://github.com/confluentinc/kafka-streams-examples/blob/5.1.0-post/src/main/java/io/confluent/examples/streams/UserRegionLambdaExample.java

Version is not final, still need to add doc, but looking for initial feedback.


99-not-out:
Remove the licence, as it is covered by the root licence (BSD-3-Clause)

examples/user-events/dev/system.clj
```diff
@@ -0,0 +1,66 @@
...
+(defn application-id
+  "Takes an application config and returns an `application.id`."
+  [app-config]
+  (get app-config "application.id"))
+
+(defn destroy-state-stores
+  "Takes an application config and deletes local files associated with
+  internal state."
+  [app-config]
+  (sh "rm" "-rf" (str "/tmp/kafka-streams/" (application-id app-config)))
```
99-not-out: Although this is no doubt useful, `rm -rf` is one of those Unix commands which can cause untold misery. I would be inclined to remove this fn from the example code as its going into the public domain ...

examples/user-events/project.clj
```diff
@@ -0,0 +1,16 @@
+(defproject user-events "0.1.0-SNAPSHOT"
+  :description "Demonstrates group-by operations and aggregations on KTable by computing
+  user-count per region"
+  :url "http://example.com/FIXME"
+  :license {:name "EPL-2.0 OR GPL-2.0-or-later WITH Classpath-exception-2.0"
```
99-not-out: pls remove

examples/user-events/project.clj
```diff
@@ -0,0 +1,16 @@
+(defproject user-events "0.1.0-SNAPSHOT"
+  :description "Demonstrates group-by operations and aggregations on KTable by computing
+  user-count per region"
+  :url "http://example.com/FIXME"
```
99-not-out: should probs be the jackdaw git url, or this part of the repo's url

99-not-out:
The actual topology looks good - we should probs jus remove the tests for now, and maybe add some "test machine" ones as a follow up

examples/user-events/dev/system.clj
astelvida: I agree on this one, I was just following the pattern in the other examples (see word-count). Best have this change applied in other parts as well?

examples/user-region/project.clj
```diff
@@ -0,0 +1,12 @@
+(defproject user-region "0.1.0-SNAPSHOT"
```
creese: I would prefer to stick with deps.edn like we do in the other examples.

examples/user-region/src/user_region/core.clj
```diff
@@ -0,0 +1,131 @@
+(ns user-region.core
```
creese: The usage of "core" is somewhat archaic. I would just call this `user-region`.

examples/user-region/src/user_region/core.clj
```diff
@@ -0,0 +1,131 @@
...
+  "Demonstrates group-by operations and aggregations on KTable.
+
+  In this specific example we compute the user count per geo-region
+  from a KTable that contains <user, region> information."
```
creese: We could link to the Java example here, if you want.

examples/user-region/CHANGELOG.md
```diff
@@ -0,0 +1,24 @@
...
+### Fixed
+- Fixed widget maker to keep working when daylight savings switches over.
+
+## 0.1.0 - 2019-02-05
+### Added
+- Files from the new template.
+- Widget maker public API - `make-widget-sync`.
+
+[Unreleased]: https://github.com/your-name/user-events/compare/0.1.1...HEAD
+[0.1.1]: https://github.com/your-name/user-events/compare/0.1.0...0.1.1
```
creese: Might want to remove this since its just a stub.

examples/user-region/src/user_region/core.clj
```diff
@@ -0,0 +1,131 @@
...
+  (let [user-region-table (j/ktable builder (topic-config "user-region"))
+
+        region-count (-> user-region-table
+                         (j/group-by (fn [[_ v]] (vector v v))
+                                     (topic-config nil
+                                                   (Serdes/String)
+                                                   (Serdes/String)))
+                         (j/count))]
+
+    ;; Aggregate the user counts of by region
```
creese: The aggregation happens in an earlier step. This just converts the table to a record stream and publishes to output.

examples/user-region/src/user_region/core.clj
```diff
@@ -0,0 +1,131 @@
...
+  a Kafka topic called `region-user-count`. Returns a topology builder."
+  [builder]
+  (let [user-region-table (j/ktable builder (topic-config "user-region"))
+
+        region-count (-> user-region-table
+                         (j/group-by (fn [[_ v]] (vector v v))
+                                     (topic-config nil
+                                                   (Serdes/String)
+                                                   (Serdes/String)))
+                         (j/count))]
```
creese: You might be missing a `j/filter` here.

examples/user-region/src/user_region/core.clj
```diff
@@ -0,0 +1,131 @@
...
+
+
+(defn topology-builder
+  "Reads from a Kafka topic called `user-region` and writes these to
+  a Kafka topic called `region-user-count`. Returns a topology builder."
+  [builder]
+  (let [user-region-table (j/ktable builder (topic-config "user-region"))
+
+        region-count (-> user-region-table
+                         (j/group-by (fn [[_ v]] (vector v v))
```
creese: The KeyValueMapper can also be written as `(fn [[_ v]] [v v])`.

examples/user-events/dev/system.clj
creese: @99-not-out How do you suggest we delete the internal state?

examples/user-region/src/user_region/core.clj
astelvida: I preferred not using the filter because this way it makes the changes more obvious in the output, especially if you run things one by one

mattford63:
Closing due to age

PR #52, opened 2019-02-05
add left-join-example

astelvida:
We show an example of performing a left join op between a K-stream and K-table. More specifically we enrich an incoming stream of events with side data, and then compute aggregations based on the enriched stream.

Inspired by: https://www.confluent.io/blog/distributed-real-time-joins-and-aggregations-on-user-activity-events-using-kafka-streams/

Version not final, looking for initial feedback. Thanks! 🚀

### DIFF
--- a/examples/user-events-region/project.clj
+++ b/examples/user-events-region/project.clj
@@ -1,0 +1,15 @@
+(defproject user-events-region "0.1.0-SNAPSHOT"
+            :description "FIXME: write description"
+            :url "http://example.com/FIXME"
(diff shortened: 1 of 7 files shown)


examples/user-events-region/project.clj
```diff
@@ -0,0 +1,15 @@
+(defproject user-events-region "0.1.0-SNAPSHOT"
+            :description "FIXME: write description"
+            :url "http://example.com/FIXME"
+            :license {:name "EPL-2.0 OR GPL-2.0-or-later WITH Classpath-exception-2.0"
```
99-not-out: Strip the licence

mattford63:
Closing due to age.

PR #53, opened 2019-02-05
[RFC] Add space for slide-deck docs

cddr:
It would be cool to publish the slides I used to explain the test machine somewhere. I tried publishing it from the google doc but people from outside Funding Circle were unable to access it. @creese wanna add yours from the conj in here too?


creese:
I don't think this repo is the right place for this. Let's try to find another location to host.

AndreaCrotti:
Slides without the video of the person presenting are not super useful imho in most cases (and also a pptx in a code repo feels a bit weird, not everyone can even open them).

But if we could extract something from there and just add some docs it would be great!

PR #54, opened 2019-02-06
Cleanup unused dependecies

AndreaCrotti:
- cleanup unused dependencies
- move codox and git-version plugin to dev profile

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -20,10 +20,6 @@
                  [org.clojure/tools.logging "0.4.1"]
                  [org.clojure/core.cache "0.7.2"]]
 
-  :plugins [[lein-codox "0.10.3"]
-            [lein-environ "1.1.0"]
-            [me.arrdem/lein-git-version "2.0.8"]]
-
   :aot [jackdaw.serdes.fn-impl]
 
   :git-version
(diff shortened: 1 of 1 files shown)


project.clj
```diff
@@ -5,18 +5,12 @@
                  [clj-time "0.15.1"]
                  [org.clojure/core.async "0.4.490"]
                  [danlentz/clj-uuid "0.1.7"]
-                 [environ "1.1.0"]
```
creese: Please coordinate with @cddr on https://github.com/FundingCircle/jackdaw/pull/55. We might need this.

cddr: That PR doesn't require `environ`. It reads the environment variables via aero in https://github.com/FundingCircle/jackdaw/pull/55/files#diff-09f64f8bbdb4c2152d66c619809f4e3c

creese: Waiting on this:
https://github.com/FundingCircle/jackdaw/pull/55/files#r254849841

AndreaCrotti: I don't think it should be waiting for that PR, it can just be added back in that PR if @cddr switches from aero to environ.

PR #55, opened 2019-02-07
Provide services required by tests in docker-compose

cddr:
## TLDR;

```
docker-compose up -d
lein test
```

## More Detail

The test-suite has some integration tests that assume the presence of a few of
the services you might expect. Currently the services required by the tests
are:-

 - zookeeper
 - kafka
 - schema-registry
 - rest-proxy

As developers, we have a veritable feast of options available to us about
exactly how these services are made available to the test-suite. For example,
your OS probably has a few mechanisms for doing it. Confluent has another.
Docker, nomad, mesosphere....You get the point.

Each has pros and cons and developers optimizing for different work-flows can
reasonably disagree about best practices for their particular scenario.

Hopefully any debate about these scenarios ends up as examples in the wiki with
a discussion of the trade-offs so that we can all understand them a bit better
and make better choices.

In the realm of automated testing scenarios, despite it's rocky past and
uncertain future, `docker-compose` right now gives us IMO the easiest way of
defining and managing a collection of services to be launched as part of a test
no matter the "testing platform" used (e.g. circleci, travis, codebuild).

Some people might find benefit in using it in their dev environments (e.g. as a
final step before submitting a pull request), but the primary reason for it's
inclusion is to ensure test results are reproducible independently of the host
environment. Discussion about how to support any particular workflow seems
likely to generate lots of heat and not much light.


README.md
```diff
@@ -20,6 +20,22 @@ To use the latest release, add the following to your project:
...
 
+## Testing
+
+The included docker-compose file (adapted from confluent's ["all-in-one"](https://github.com/confluentinc/cp-docker-images/blob/5.1.0-post/examples/cp-all-in-one/docker-compose.yml) compose file)
+provides all the services required to run the project's tests. This allows the tests
+to be run as follows. You'll need up-to-date versions of docker, docker-compose and lein
+
+```
+docker-compose up -d
+docker run \
```
AndreaCrotti: This might not be a very friendly way to run tests imho.

Can't we just do like in other projects where we simply do something like
`lein with-profile docker test` for example, so you still run everything locally but simply connecting to the docker containers?

docker-compose.yml
```diff
@@ -0,0 +1,56 @@
+version: '3'
+services:
+  zookeeper:
+    image: confluentinc/cp-zookeeper:5.1.0
+    hostname: zookeeper
+    container_name: zookeeper
+    expose:
+      - "2181"
```
AndreaCrotti: To be able to run tests as I was mentioning just above we just need to forward a few ports here, as long as they are not default ports we should be fine and not at risk of colliding with the host kafka / zookeeper/etc

project.clj
```diff
@@ -65,7 +65,8 @@
               ["dev"]
 
               :injections [(require 'io.aviso.logging.setup)]
-              :dependencies [[io.aviso/logging "0.3.1"]
+              :dependencies [[aero "1.1.3"]
```
creese: Can we use `environ` instead? Aero encourages a "two config file" pattern which is a bit verbose.

cddr: How does aero encourage a two config file pattern?

    When config is left out of source code control it festers
    and diverges from the code base. Better to keep a single 
    config file in source code control.    

-- https://github.com/juxt/aero#allow-config-to-be-stored-in-the-source-code-repository-

.circleci/config.yml
```diff
@@ -89,7 +69,26 @@ jobs:
...
+
+            docker cp /home/circleci/.m2 data:/home/circleci/
+            docker cp /home/circleci/jackdaw data:/home/circleci/
+
+            docker run \
+              --env-file .env.circle \
+              --workdir /home/circleci/jackdaw \
+              --volumes-from data \
+              --network jackdaw_default \
+              circleci/clojure:lein-2.8.1 lein test
```
creese: The CircleCI documentation recommends the approach used in master. Please see ["Using Multiple Docker Images"](https://circleci.com/docs/2.0/executor-types/#using-multiple-docker-images). I suggest we stay closer to that. 

https://github.com/FundingCircle/jackdaw/pull/66 shows how we can extend this to test against multiple versions of the Confluent Platform: https://github.com/FundingCircle/jackdaw/pull/66. The tests run in parallel.

docker-compose.yml
```diff
@@ -0,0 +1,56 @@
...
+      - schema-registry
+    expose:
+      - "8082"
+    hostname: rest-proxy
+    container_name: rest-proxy
+    environment:
+      KAFKA_REST_HOST_NAME: rest-proxy
+      KAFKA_REST_BOOTSTRAP_SERVERS: 'broker:9092'
+      KAFKA_REST_LISTENERS: "http://0.0.0.0:8082"
+      KAFKA_REST_SCHEMA_REGISTRY_URL: 'http://schema-registry:8081'
```
creese: I think this adds tremendous value, but I would also like to keep Jackdaw agnostic on how users run Kafka for local testing. Can we move this to documentation in `doc/`. We could include this docker-compose file as a code snippet for the user to cut-and-paste out.

cddr: Shall we `.gitignore` it so that folks can use it for testing jackdaw itself without seeing it show up as a diff?

cddr: Ah no need (from the git docs)...

    Patterns which are specific to a particular repository but which
    do not need to be shared with other related repositories (e.g., 
    auxiliary files that live inside the repository but are specific 
    to one user’s workflow) should go into the $GIT_DIR/info/exclude
    file.

-- https://git-scm.com/docs/gitignore

PR #60, opened 2019-02-08
Use manifold for all async in the test-machine

cddr:
Replaces all use of core.async with the equivalent manifold construct. Basically, this means..

  - replacing `channel`s with `stream`s
  - replacing `go-loop`s with `d/loops`s
  - wrapping the bodies of the `go-loops` into `d/chain` handlers

This means we can remove the unnecessary dependency on core.async which can cause problems depending on the downstream application's other dependencies.

Also fixed a startup bug in the kstreams fixture that results in the possibility of tests being commenced without the app under test having reached the started state.

### DIFF
--- a/src/jackdaw/test/fixtures.clj
+++ b/src/jackdaw/test/fixtures.clj
@@ -83,16 +83,16 @@
 ;;; kstream-fixture ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (defn- set-started
-  [app-id started]
+  [app-id started?]
   (reify KafkaStreams$StateListener
     (onChange [_ new-state old-state]
       (log/infof "process %s changed state from %s -> %s"
                  app-id
                  (.name old-state)
                  (.name new-state))
-      (when-not @started
+      (when-not (realized? started?)
         (when (.isRunning new-state)
-          (reset! started true))))))
+          (deliver started? true))))))
 
 (defn- set-error
   [error]
@@ -113,14 +113,15 @@
     (let [builder (k/streams-builder)
           stream (k/kafka-streams (topology builder) config)
           error (atom nil)
-          started (atom false)]
+          started? (promise)]
 
       (.setUncaughtExceptionHandler stream (set-error error))
-      (.setStateListener stream (set-started (get config "application.id") started))
+      (.setStateListener stream (set-started (get config "application.id") started?))
 
       (k/start stream)
 
-      (when @started)
+      (when @started?
+        (log/info "commencing test function"))
 
       (try
         (t)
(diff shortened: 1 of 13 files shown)


src/jackdaw/test/fixtures.clj
```diff
@@ -111,14 +111,15 @@
...
-          started (atom false)]
+          started? (promise)]
 
       (.setUncaughtExceptionHandler stream (set-error error))
-      (.setStateListener stream (set-started (get config "application.id") started))
+      (.setStateListener stream (set-started (get config "application.id") started?))
 
       (k/start stream)
 
-      (when @started)
```
cddr: This was the bug. At one point (pre-jackdaw), `started` was a promise that blocked until the app under test was ready. When it got changed to an atom, most of the time it would be ok but if the app is slow to reach the "running" state, the test might start firing requests before the app is able to handle them.

src/jackdaw/test/journal.clj
```diff
@@ -76,21 +78,28 @@
     (throw (ex-info "no message stream to journal:" machine)))
 
   (let [{:keys [messages]} (:consumer machine)]
-    (async/go-loop [record (async/<! messages)]
-      (when record
-        (journal-result machine record)
-        (recur (async/<! messages))))))
+    (d/loop [record (s/take! messages)]
+      (d/chain record
```
kachayev: @cddr Here and further... You can use `d/chain'` instead of `d/chain` any time you're sure that the argument would be a manifold deferred to avoid unnecessary runtime coercion.

src/jackdaw/test/journal.clj
```diff
@@ -76,21 +78,28 @@
...
+               (fn [record]
+                 (when-not @stop?
```
kachayev: @cddr Doing this you can cancel `d/loop` (stop it) only when there's a new message you can read from the stream. Usually, it's better to check closing conditions periodically on idle. The easiest way would be `take!` with the timeout. E.g.

```clojure
(d/loop [record (s/try-take! message ::drained 1e3 ::timeout)]
  (cond
     (true? @stop)
     <..>

     (identical? ::drained message)
     <..> ;; no more messages

     (identical? ::timeout message)
     (d/recur <..>)))
```

src/jackdaw/test/transports/kafka.clj
```diff
@@ -156,30 +158,32 @@
...
-                                   (recur (async/<! messages)))
-         serialization-error   (do (deliver ack {:error :serialization-error
-                                                 :message (.getMessage serialization-error)})
-                                   (recur (async/<! messages)))
-         :else (do
-                 (.close producer)
-                 (log/infof "stopped kafka producer: "
-                            (select-keys kafka-config ["bootstrap.servers" "group.id"])))))
+     (d/loop [message (s/take! messages)]
+       (d/chain (d/future message)
```
kachayev: @cddr  `message` here is already a manifold deferred, why do you need to wrap it into a `future`?

PR #63, opened 2019-02-08
Auto-deploy to clojars for pushed tags matching /^\d+.\d+.\d+/

### LINKED ISSUE #35: Automatically push release to Clojars when a release is created
Releasing the library manually is painful, and potentially error prone - we should have the tooling do it automatically when a release is created in Git hub (or something like that)

cddr:
This PR adds a circleci workflow for automatically pushing tagged releases to clojars. The workflow only runs for tags, not branches.

To actually trigger a release therefore, there are a couple of options...

  - use the github release UI to create a release that when published, will commit a corresponding tag to the repo
  - at the command line, something like `git tag -a v0.5.2 -m "Release v0.5.2"`

## TODO

 - [ ] Add clojars credentials to the private project settings (needs project admin)
 - [x] Update branch protection to check for "test" step instead of "build" step (needs project admin)

Maybe resolves #35

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,6 +35,11 @@ references:
       - image: circleci/clojure:lein-2.8.1
     working_directory: ~/jackdaw
 
+  deploy_config: &deploy_config
+    docker:
+      - image: circleci/clojure:lein-2.8.1
+    working_directory: /home/circleci/jackdaw
+
   test_config: &test_config
     docker:
       - image: circleci/clojure:lein-2.8.1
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -70,6 +75,18 @@ jobs:
           key: v1-jackdaw-{{ .Branch }}-{{ .Revision }}
           paths:
             - .
+  checkout_tags:
+    <<: *build_config
+    steps:
+      - *restore_repo
+      - run: git config --global url."https://github.com/".insteadOf "git@github.com:"
```
cddr: This little dance is required because
  a. the cached repo doesn't include the tags so the subsequent `checkout` step fails.
  b. the clone url for the repo uses `git:github...` so the fetch triggers an interactive ssh host check so we want to switch to using https

.circleci/config.yml
```diff
@@ -93,16 +110,41 @@ jobs:
...
+              only: /.*/
+            branches:
+              ignore: /.*/
+      - deploy:
+          context: org-global
+          requires:
+            - checkout_tags
+          filters:
+            tags:
+              only: /^v.*/
```
creese: I would prefer to keep the "v" out of the tag. Can we release on tags matching `x.y.z`?  (For instance, anything matching `^\d+\.\d+\.\d+$`.)

cddr: Sure. But it was useful while testing to match version strings with arbitrary suffixes. Otherwise in order to test this workflow, you need to make an official looking release.

I think it would be useful to match `/^\d+\.\d+\.\d+.-/` to support this use-case and also allow for release candidates like "1.2.3-rc1"

creese: Should we exclude `SNAPSHOT` from the list of allowed suffixes?

PR #64, opened 2019-02-09
Backfill unit tests for jackdaw.admin

cddr:
I looked into the admin ns because it had quite low test coverage. Think I might have found a bug and hopefully improved test coverage a bit with some of the easier tests. The remaining functions will be a bit more difficult because the MockAdminClient doesn't cover all the methods we use on the AdminClient.

### DIFF
--- a/src/jackdaw/admin.clj
+++ b/src/jackdaw/admin.clj
@@ -23,33 +23,26 @@
 
 (def client-impl
   {:alter-topics* (fn [this topics]
-                   (d/future
-                     (->> (.alterConfigs this topics)
-                          .all deref)))
+                    (d/future
+                      @(.all (.alterConfigs this topics))))
    :create-topics* (fn [this topics]
                     (d/future
-                      (->> (.createTopics this topics)
-                           .all deref)))
+                      @(.all (.createTopics this topics))))
    :delete-topics*  (fn [this topics]
                       (d/future
-                        (->> (.deleteTopics this topics)
-                             .all deref)))
+                        @(.all (.deleteTopics this topics))))
    :describe-topics* (fn [this topics]
                        (d/future
-                         (->> (.describeTopics this topics (DescribeTopicsOptions.))
-                              .all deref)))
+                         @(.all (.describeTopics this topics (DescribeTopicsOptions.)))))
    :describe-configs* (fn [this configs]
                         (d/future
-                          (->> (.describeConfigs this configs (DescribeConfigsOptions.))
-                               .all .get)))
+                          @(.all (.describeConfigs this configs (DescribeConfigsOptions.)))))
    :describe-cluster* (fn [this]
-                       (d/future
-                         (->> (.describeCluster this (DescribeClusterOptions.))
-                              jd/datafy)))
+                        (d/future
+                          (jd/datafy (.describeCluster this (DescribeClusterOptions.)))))
    :list-topics* (fn [this]
-                  (d/future
-                    (->> (.listTopics this)
-                         .names deref)))})
+                   (d/future
+                     @(.names (.listTopics this))))})
 
 (extend AdminClient
   Client
--- a/src/jackdaw/admin.clj
+++ b/src/jackdaw/admin.clj
@@ -157,7 +157,7 @@
   This can be used to determine if some set of newly created topics
   are healthy yet, or detect whether leader re-election has finished
   following the demise of a Kafka broker."
-  [client topics]
+  [^AdminClient client topics]
   {:pre [(client? client)
          (sequential? topics)]}
   (->> (describe-topics* client (map :topic-name topics))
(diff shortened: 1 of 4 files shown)


src/jackdaw/admin.clj
```diff
@@ -6,11 +6,54 @@
...
-  (:require [jackdaw.data :as jd])
-  (:import [org.apache.kafka.clients.admin
-            AdminClient DescribeTopicsOptions]))
-
-(set! *warn-on-reflection* true)
+  (:require
+   [jackdaw.data :as jd]
+   [manifold.deferred :as d])
+  (:import [org.apache.kafka.clients.admin AdminClient
+            DescribeTopicsOptions DescribeClusterOptions DescribeConfigsOptions]))
```
rockBreaker: I don't think DescribeTopicsOptions is used.
```suggestion
            DescribeClusterOptions DescribeConfigsOptions]))
```

test/jackdaw/admin_test.clj
```diff
@@ -0,0 +1,163 @@
+(ns jackdaw.admin-test
+  (:require
+   [clojure.test :refer :all]
+   [clojure.spec.alpha :as s]
```
rockBreaker: I don't think clojure.spec is used 
```suggestion
 
```

test/jackdaw/admin_test.clj
```diff
@@ -0,0 +1,163 @@
...
+   [jackdaw.admin :as admin]
+   [jackdaw.data :as data]
+   [manifold.deferred :as d])
+  (:import
+   (org.apache.kafka.common Node KafkaFuture)
+   (org.apache.kafka.clients.admin MockAdminClient
+    DescribeTopicsOptions DescribeClusterOptions DescribeConfigsOptions)))
```
rockBreaker: I don't think  DescribeTopicsOptions DescribeClusterOptions DescribeConfigsOptions are used 
```suggestion
   )))
```

src/jackdaw/admin.clj
```diff
@@ -6,11 +6,54 @@
...
+(defprotocol Client
+  (alter-topics* [this topics])
+  (create-topics* [this topics])
+  (delete-topics* [this topics])
+  (describe-topics* [this topics])
+  (describe-configs* [this configs])
+  (describe-cluster* [this])
+  (list-topics* [this]))
+
+(def client-impl
```
rockBreaker: It might be nice to follow the same convention for this def.

Should it be `->` or `->>`

Should it be on one line or multiple lines?
e.g
`(-> (.alterConfigs this topics) .all deref)`
or
``` 
(-> (.describeConfigs this configs (DescribeConfigsOptions.))
                              .all .get)
```

test/jackdaw/admin_test.clj
```diff
@@ -0,0 +1,163 @@
...
+    (fn [client]
+      (admin/create-topics! client (vals test-topics))
+      (is (set= (map #(select-keys % [:topic-name]) (vals test-topics))
+                (admin/list-topics client))))))
+
+(deftest test-topic-exists?
+  (with-mock-admin-client test-cluster
+    (fn [client]
+      (admin/create-topics! client (vals test-topics))
+
```
rockBreaker: Could remove the whitespace here

src/jackdaw/admin.clj
cddr: Agree it would be nice to get some consistency here.  I've gone for two lines. The first being the API call while the second "extracts" the value in a way that might differ for each type of command (mostly it's just `.all .deref` but in some cases the logical equivalent is a bit different like the describeCluster which returns a map of futures which are forced by the call to `datafy`).

```
(->> (.apiCall this args (Options.))
         .extract-result)
```

The idea here is to perform the bare minimum API call and extract the result without any kind of transformation

src/jackdaw/admin.clj
cddr: I made the client-impl use it for consistency with the other similar API calls. It seems like the MockAdminClient is more likely to support the variant that supplies the options so explicitly supplying those has benefits from a testing point of view.

src/jackdaw/admin.clj
```diff
@@ -6,11 +6,54 @@
...
+  {:alter-topics* (fn [this topics]
+                   (d/future
+                     (->> (.alterConfigs this topics)
+                          .all deref)))
```
99-not-out: On style, is `(deref (.all (.alterConfigs this topics)))` or `@(.all (.alterConfigs this topics))` actually clearer here? I _personally_ find it easier than the threading macro version (but I have never got on with the threading macro anyway ...).

src/jackdaw/admin.clj
rockBreaker: Ah apologies didn't see! Ill resolve.

src/jackdaw/admin.clj
rockBreaker: +1

PR #67, opened 2019-02-11
Report test coverage to codecov

cddr:
Extracts the cloverage part out of #16. Back when that was first created, the ci pipeline was a bit more complicated. Much simpler to run code coverage as part of the test step now.


AndreaCrotti:
Nice one!
You could add a codecov badge in the README, close to the CircleCI, so everyone would know more easily we have code coverage, but otherwise looks good!

(sadly both badges will always point to master of course but that's still good to have imho)

PR #74, opened 2019-02-15
Parameterize test environment variables

cddr:
Parameterize the test-config so that the tests can be run against any host/port, not just localhost.

### DIFF
--- a/test/jackdaw/serdes/avro/integration_test.clj
+++ b/test/jackdaw/serdes/avro/integration_test.clj
@@ -9,13 +9,17 @@
             [jackdaw.data :as jd]
             [jackdaw.serdes.avro :as avro]
             [jackdaw.serdes.avro.schema-registry :as reg]
-            [jackdaw.test.fixtures :as fix])
+            [jackdaw.test.fixtures :as fix]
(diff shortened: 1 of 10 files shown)


test/jackdaw/serdes/avro/integration_test.clj
```diff
@@ -9,13 +9,17 @@
...
+            [jackdaw.test-config :refer [test-config]])
   (:import [org.apache.avro Schema$Parser]
            [org.apache.avro.generic GenericData$Record]
            [org.apache.kafka.common.serialization Serde Serdes]))
 
-(def +real-schema-registry-url+
-  "http://localhost:8081")
+(defn ^:private real-schema-registry-url
+  []
+  (format "http://%s:%s"
```
DaveWM: This code is repeated quite a lot (substituting `:schema-registry` for other keywords), could it be extracted into a function in the `jackdaw.test-config` ns? Maybe something like:
```clojure
(defn get-url [service]
  (format "%s:%s"
          (get-in (test-config) [service :host])
          (get-in (test-config) [service :port])))
```
Then the code here would become:
```clojure
(jackdaw.test-config/get-url :schema-registry)
```

PR #71, opened 2019-02-11
Serde Resolver

99-not-out:
Adds a "serve resolver" which given a serdes config with `:key-serde`  and `:value-serde` keys in it, replaces the values for those keys with the actual  implementations for the required serde. Accepts options to specify any extra information which may be needed to create a serde, e.g. the avro serdes needs to know the  schema registry URL.

The new metadata format looks like:
```
(def +topic-metadata+
  {"input"
   {:topic-name "input"
    :partition-count 1
    :replication-factor 1
    :key-serde {:serde-qualified-keyword :jackdaw.serdes/string-serde}
    :value-serde {:serde-qualified-keyword :jackdaw.serdes.avro.confluent/serde
                  :schema-filename "value-schema.json"
                  :key? false}}

   "output"
   {:topic-name "output"
    :partition-count 1
    :replication-factor 1
    :key-serde {:serde-qualified-keyword :jackdaw.serdes/string-serde}
    :value-serde {:serde-qualified-keyword :jackdaw.serdes.avro.confluent/serde
                  :schema-filename "value-schema.json"
                  :key? false}}})
```

The resolver namespace contains a function called `serde-resolver` that returns a function of arity one which takes a **serde config** (as opposed to a topic config) and resolves the serde. Only one serde is resolved at a time.

This is the serde spec:
```
(s/def ::serde-qualified-keyword qualified-keyword?)
(s/def ::schema string?)
(s/def ::schema-filename string?)
(s/def ::key? boolean?)

(s/def ::jackdaw.serde/serde
  (s/keys :req-un [::serde-qualified-keyword]
          :opt-un [::schema
                   ::schema-filename
                   ::key?]))
```

To update some topic "metadata" with serdes resolved, there is still a bit of glue code (but not very much!):
```
(def serde-resolver
  (partial resolver/serde-resolver :schema-registry-url "http://localhost:8081"))

(def topic-metadata
  (memoize (fn []
             (fmap #(assoc % :key-serde ((serde-resolver) (:key-serde %))
                           :value-serde ((serde-resolver) (:value-serde %)))
                   +topic-metadata+))))
```

There are some examples added also in this PR.

PR #83, opened 2019-02-15
Add integration tests for client namespace

cddr:
Adds a bunch of tests for the client namespace. This PR covers the following aspects.

  - basic consumer/producer creation and teardown
  - send!, produce!
  - topic subscription and assignment
  - topic partition interrogation

There's still a bit more that remains without tests like offset setting/getting but these features are used by the test-machine's own test-suite.

### DIFF
--- a/src/jackdaw/client.clj
+++ b/src/jackdaw/client.clj
@@ -22,6 +22,7 @@
            org.apache.kafka.common.serialization.Serde))
 
 (set! *warn-on-reflection* true)
+(declare assignment)
 
 ;;;; Producer
 
@@ -119,11 +120,6 @@
   [^KafkaConsumer consumer]
   (.subscription consumer))
 
-(defn assignment
-  "Return the assigned topics and partitions of a consumer."
-  [^KafkaConsumer consumer]
-  (map jd/datafy (.assignment consumer)))
-
 (defn subscribe
   "Subscribe a consumer to the specified topics.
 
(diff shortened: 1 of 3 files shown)


src/jackdaw/client.clj
```diff
@@ -305,3 +301,8 @@
                         (map (fn [^PartitionInfo x]
                                (TopicPartition. (.topic x) (.partition x)))))]
     (apply assign consumer partitions)))
+
+(defn assignment
+  "Return the assigned topics and partitions of a consumer."
+  [^KafkaConsumer consumer]
+  (map jd/datafy (.assignment consumer)))
```
creese: Why did you move this to the bottom?

test/jackdaw/client_test.clj
```diff
@@ -1,12 +1,85 @@
...
+            [jackdaw.test.serde :as serde]
+            [jackdaw.data :as data])
   (:import [java.util.concurrent LinkedBlockingQueue TimeUnit]
            java.time.Duration
            [org.apache.kafka.clients.consumer Consumer ConsumerRecord ConsumerRecords]
            [org.apache.kafka.clients.producer Producer]
            org.apache.kafka.common.TopicPartition))
 
+(def foo-topic
+  (serde/resolver {:topic-name "foo"
```
creese: Not a blocker, but we should refactor this to use `jackdaw.serdes.resolver/serde-resolver` after https://github.com/FundingCircle/jackdaw/pull/71 is merged.

src/jackdaw/client.clj
kidpollo: dont exactly recall doing that :P was not intentional

kidpollo:
YAY THANKS!

src/jackdaw/client.clj
cddr: To keep it together with the function that lets you set the assignments.

PR #90, opened 2019-02-24
Test word count example

cddr:
This PR adds an example app (complete with e2e tests) to the jackdaw test suite. The example uses a Clojure port of the word-count demo from the [confluent kafka examples](https://github.com/confluentinc/kafka-streams-examples).

In support of this addition, the PR also includes....

  * a new fixture for resetting application state
  * a fix to the `k/count` fspec so that calls to `count` can fall back to the default application configured serdes
  * a fix to the test-machine so that it doesn't try to execute the first command if the test-sequence is empty
  * a fix to the kafka transport in the test-machine to ensure that journal output is identified by the topic id specified in the mapping supplied to the test-machine

Happy to split these into separate PRs if that's desirable.


src/jackdaw/streams/specs.clj
```diff
@@ -276,7 +276,7 @@
 
 (s/fdef k/count
         :args (s/cat :kgrouped ::kgroupedstream-or-kgroupedtable
-                     :topic-config ::topic-config)
+                     :topic-config (s/? ::topic-config))
```
cddr: There might be a few more cases like this. Created #91 to investigate.

test/jackdaw/word_count_test.clj
```diff
@@ -0,0 +1,160 @@
+(ns jackdaw.word-count-test
```
creese: Can we move this over to the [example](https://github.com/FundingCircle/jackdaw/blob/master/examples/word-count/test/word_count_test.clj)? I think that might be a little more DRY, and the example could use a little TM love.

test/jackdaw/word_count_test.clj
```diff
@@ -0,0 +1,160 @@
...
+   "default.key.serde" "org.apache.kafka.common.serialization.Serdes$StringSerde"
+   "default.value.serde" "org.apache.kafka.common.serialization.Serdes$StringSerde"))
+
+(def test-config
+  (assoc broker-config
+    "group.id" "word-count-test"))
+
+;; Now we're ready to build a test
+
+(defn write-lines
```
creese: This function doesn't actually write lines, does it? Should we maybe name it after the data returned, for example, `write-commands`? 

The same could be said of `jackdaw.word-count-test/watch-for-output` below.

test/jackdaw/word_count_test.clj
```diff
@@ -0,0 +1,160 @@
...
+;;   - Write the counts to the output topic
+
+(defn parse-line
+  [line]
+  (let [line (-> (.toLowerCase line (java.util.Locale/getDefault)))]
+    (->> (.split line "\\W+")
+         (into []))))
+
+(defn word-count
+  [in out]
```
creese: If we switch to a "topic metadata" format like we use in the [serdes example](https://github.com/FundingCircle/jackdaw/blob/master/examples/serdes/src/serdes.clj#L129-L159), this the function could be made to take one argument.

test/jackdaw/word_count_test.clj
```diff
@@ -0,0 +1,160 @@
...
+  [word]
+  [[:watch (fn [journal]
+             (some #(= word (:key %))
+                   (get-in journal [:topics :output]))) 2000]])
+
+(deftest test-word-count-demo
+  (fix/with-fixtures [(fix/topic-fixture broker-config word-count-topics)
+                      (fix/reset-application-fixture app-config)
+                      (fix/kstream-fixture {:topology (word-count input output)
+                                            :config app-config})]
```
creese: I think we should combine these into a single fixture.

src/jackdaw/test/fixtures.clj
```diff
@@ -55,6 +57,42 @@
...
+(defn reset-application-fixture [app-config]
+  (fn [t]
+    (let [rt (StreamsResetter.)
+          app-id (get app-config "application.id")
+          app-state (format "%s/%s"
+                            (or (get app-config "state.dir")
+                                "/tmp/kafka-streams")
+                            app-id)]
+      (when (.exists (io/as-file app-state))
+        (delete-files-recursively app-state))
```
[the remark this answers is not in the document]
cddr: The app should not be running at the point of this function being called. But @gphilipp thought this might be better as an option to the kstream-fixture that could just use the `.cleanUp` method of the kstream it has just created (but not yet started). A similar option on the `topic-fixture` could be used to trigger topic-deletion before re-creation.

test/jackdaw/word_count_test.clj
cddr: I started with that but wasn't sure what the examples part of repo was trying to achieve. Is it an interactive demo/tutorial? Or a runnable test. It seems hard to keep it simple when it tries to be both. I think it's useful to have one complete e2e app and test as part of the test-suite.

test/jackdaw/word_count_test.clj
cddr: These are just private test helpers so I don't think the names are hugely important but I think that `write-commands` is worse. I tried to choose names for how they will eventually be used in the context of a test. The "lines" bit is intended to make the language of the test match the language of problem being solved.

```
      (let [{:keys [results journal]} (->> (concat
                                            (write-lines)
                                            (watch-for-output "understand"))
                                           (jd.test/run-test machine))
```

Within the context of the `concat`, tests will often use helpers to build up commands. It would get pretty tedious if every function had the word "commands" in it.

src/jackdaw/test/fixtures.clj
```diff
@@ -41,19 +45,83 @@
...
 
-  ([kafka-config topic-config timeout-ms]
+  ([kafka-config topic-config opts]
    (fn [t]
-     (with-open [client (AdminClient/create kafka-config)]
-       (-> (create-topics client kafka-config topic-config)
-           (.get timeout-ms java.util.concurrent.TimeUnit/MILLISECONDS))
-       (log/info "topic-fixture: created topics: " (keys topic-config))
-       (t)))))
+     (let [{:keys [timeout-ms delete-first?]} (parse-opts opts)]
```
cddr: `parse-opts` is introduced in order to maintain compatibility with the existing version. Callers that provide an integer for timeout-ms will continue to behave the same way.

src/jackdaw/test/fixtures.clj
```diff
@@ -108,13 +176,16 @@
 
    `compose-fixtures` or `join-fixtures` may be used to build fixtures combine
    topologies"
-  [{:keys [topology config]}]
+  [{:keys [topology config cleanup-first?]}]
```
cddr: Should be additive only

test/jackdaw/client_test.clj
```diff
@@ -104,7 +104,7 @@
 
 
 (deftest ^:integration send!-test
-  (fix/with-fixtures [(fix/topic-fixture (broker-config) test-topics 1000)]
+  (fix/with-fixtures [(fix/topic-fixture (broker-config) test-topics {:timeout-ms 1000})]
```
cddr: This is the new way to express a timeout if you want something different from the default. The old way still works but should be considered deprecated.

PR #92, opened 2019-02-25
Add a signing key to get the auto deploy pipeline unblocked

### LINKED ISSUE #88: Deploy Fails for Pushed Tag
```
$ #!/bin/bash -eo pipefail
lein deploy clojars

No credentials found for clojars
See `lein help deploying` for how to configure credentials to avoid prompts.
Username: Password: 
Compiling jackdaw.serdes.fn-impl
Created /home/circleci/jackdaw/target/jackdaw-0.6.0.jar
Wrote /home/circleci/jackdaw/pom.xml
Need to sign 2 files with GPG
[1/2] Signing /home/circleci/jackdaw/target/jackdaw-0.6.0.jar with GPG
gpg: directory '/home/circleci/.gnupg' created
gpg: keybox '/home/circleci/.gnupg/pubring.kbx' created
gpg: no default secret key: No secret key
gpg: signing failed: No secret key
Could not sign /home/circleci/jackdaw/target/jackdaw-0.6.0.jar
gpg: directory '/home/circleci/.gnupg' created
gpg: keybox '/home/circleci/.gnupg/pubring.kbx' created
gpg: no default secret key: No secret key
gpg: signing failed: No secret key


See `lein help gpg` for how to set up gpg.
If you don't expect people to need to verify the authorship of your jar, you
can add `:sign-releases false` to the relevant `:deploy-repositories` entry.
Exited with code 1
```

cddr:
Bit of circleci magic to get signed releases deployed to clojars for all pushed tags matching the regex `/\d+.\d+\d+/`.

Fixes #88. The behavior now is to publish a snapshot release like `n.n.n-feature-SNAPSHOT` for all passing builds. On pushed tags matching a semantic version identifier, it publishes an official release

Here's a successful test-run: https://circleci.com/workflow-run/af9f245f-8bcd-4049-9cea-3d60b81bc021


.circleci/config.yml
```diff
@@ -76,12 +76,18 @@ jobs:
           paths:
             - .
   checkout_tags:
-    <<: *build_config
+    <<: *deploy_config
```
creese: `deploy_config` and `build_config` are the same now. Can we remove one?

.circleci/fundingcirclebot.key
```diff
@@ -0,0 +1,60 @@
+-----BEGIN PGP PRIVATE KEY BLOCK-----
```
creese: Can you explain why it's okay to place this under version control?

project.clj
```diff
@@ -43,12 +43,13 @@
               :license {:name "BSD 3-clause"
                         :url "http://opensource.org/licenses/BSD-3-Clause"}
 
-              :repositories
+              :deploy-repositories
```
creese: I think we still want `"confluent"` under `:repositories`.

.circleci/config.yml
```diff
@@ -76,12 +76,18 @@ jobs:
...
       - *restore_repo
-      - run: git config --global url."https://github.com/".insteadOf "git@github.com:"
-      - run: git config --global url."https://".insteadOf "git://"
-      - run: git fetch --tags
+      # Temporary workaround for https://discuss.circleci.com/t/22437
+      - run: |
+          if [ -n "$CIRCLE_TAG" ]
+          then
+            mkdir -p ~/.ssh
+            echo 'github.com ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ==' \
```
creese: Can you explain why it's okay to place this under version control?

.circleci/config.yml
cddr: I think it's worth keeping deploy config distinct from build config as it is just a coincidence that they are the same at the moment. No reason to assume that they would remain the same forever as they serve different purposes.

.circleci/fundingcirclebot.key
cddr: This is just an unpublished key for demo purposes. We should make a real one and encrypt it.

.circleci/config.yml
cddr: This is a line we probably all have in our ~/.ssh/known_hosts but I think it's better to verify the ssh fingerprint independently if possible before adding it to known_hosts

https://stackoverflow.com/a/54090063/48641

.circleci/fundingcirclebot.key
cddr: I updated the PR with an encrypted version of a new private key. We'll need to publish this and store the private key somewhere safe.

PR #93, opened 2019-02-25
Add circleci badge

AndreaCrotti:


README.md
```diff
@@ -7,6 +7,7 @@ Jackdaw is a Clojure library for the Apache Kafka distributed streaming platform
 
 To use the latest release, add the following to your project:
 
+[![CircleCI](https://circleci.com/gh/FundingCircle/jackdaw.svg?style=svg)](https://circleci.com/gh/FundingCircle/jackdaw)
```
cddr: The list of badges looks a bit weird when prefixed with the sentence "To use the latest release, add the following to your project". Maybe we should split out the clojars one from the others?

screen shot 2019-02-25 at 9 24 43 pm

AndreaCrotti: Yes that's a good point, well I just suggest to move them all on top and remove the extra docs, I think Clojure devs can recognize the clojars badge anyway we don't really need to specify it

README.md
```diff
@@ -1,16 +1,12 @@
...
-
-Jackdaw is a Clojure library for the Apache Kafka distributed streaming platform. With Jackdaw, you can create and list topics using the AdminClient API, produce and consume records using the Producer and Consumer APIs, and create stream processing applications using the Streams API. Jackdaw also contains functions to serialize and deserialize records as JSON, EDN, and Avro, as well as functions for writing unit and integration tests.
-
-
-## Installation
-
-To use the latest release, add the following to your project:
-
 [![Clojars Project](https://img.shields.io/clojars/v/fundingcircle/jackdaw.svg)](https://clojars.org/fundingcircle/jackdaw)
+[![CircleCI](https://circleci.com/gh/FundingCircle/jackdaw.svg?style=svg)](https://circleci.com/gh/FundingCircle/jackdaw)
```
creese: Let's move the CircleCI badge to the far right and find one that matches. We may also want to replace the graphic with text while we rethink the layout.

Example:
https://github.com/facebook/react

AndreaCrotti:
Ok @creese with the latest changes it looks like this (which I think it's pretty similar to the react example):
jackdaw_badges

PR #94, opened 2019-02-25
Add guide for the streams dsl

cddr:


doc/streams-dsl.md
```diff
@@ -0,0 +1,78 @@
+# Kafka Streams DSL
```
creese: This duplicates much of what is already in the Word Count example. I'm wondering if we should improve the [README](https://github.com/FundingCircle/jackdaw/blob/master/examples/word-count/README.md) and associated [example](https://github.com/FundingCircle/jackdaw/blob/master/examples/word-count/src/word_count.clj) instead.

doc/streams-dsl.md
```diff
@@ -0,0 +1,78 @@
...
+
+If you've used the Java API, you'll be aware that the core operators are defined
+as methods on the KStream and KTable classes. In Jackdaw, we expose these
+methods as functions in the jackdaw.streams namespace with names that are
+hyphenated versions of the corresponding Java method.
+
+The [API
+docs](https://cljdoc.org/d/fundingcircle/jackdaw/CURRENT/api/jackdaw.streams)
+should be consulted for full details but the essential elements of a typical
+streams app are described below
```
creese: We could also link to the Word Count example here and remove the code below.

doc/streams-dsl.md
```diff
@@ -0,0 +1,78 @@
+# Kafka Streams DSL
+
+## Rationale
+
+The Jackdaw Streams API is a thin wrapper around the underlying Kafka Streams
+DSL. It allows the definition of kafka streams applications using idiomatic
```
creese: The official branding is "Kafka Streams", the "Kafka Streams API", and sometimes "Streams API" (but not, "DSL"). If we want to draw a distinction between that and Jackdaw, we could call it the "Java Kafka Streams API".

creese:
This might also work as a namespace doc-string in `jackdaw.streams`.

doc/streams-dsl.md
cddr: What documentation do you mean? It is referred to as a DSL in title of the streams sections in both kafka and confluent documentation.

https://docs.confluent.io/current/streams/developer-guide/dsl-api.html
https://kafka.apache.org/21/documentation/streams/developer-guide/dsl-api.html

cddr:
I'm hoping to build up a collection of docs more focused on how to use each layer of the API (together with a bit of discussion of when you might want to use it). I think the namespace doc could be a few words describing the contents (the admin ns seems like a good template).

https://cljdoc.org/d/fundingcircle/jackdaw/CURRENT/doc/readme

PR #95, opened 2019-02-25
Allow usage of timestamp extractor and reset policy in topic config

joakimmohn:
If `nil` the defaults from stream config will be used.

### DIFF
--- a/src/jackdaw/streams/interop.clj
+++ b/src/jackdaw/streams/interop.clj
@@ -27,8 +27,8 @@
 
 (set! *warn-on-reflection* true)
 
-(defn topic->consumed [{:keys [key-serde value-serde]}]
-  (Consumed/with key-serde value-serde))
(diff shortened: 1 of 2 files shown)


joakimmohn:
It looks like there's no useful type checking in the specs for the serdes in the topic config `(s/def ::key-serde any?)`

The java ks library will throw a useful exception if the wrong type is supplied: `java.lang.ClassCastException: java.base/java.lang.String cannot be cast to org.apache.kafka.streams.processor.TimestampExtractor`.

What do you think?

creese:
When making changes to the topic config, we should update the specs in `jackdaw.specs`. That will  help us decide if these belong in the topic config or if they should be new positional arguments.

creese:
> It looks like there's no useful type checking in the specs for the serdes in the topic config `(s/def ::key-serde any?)`

There is better type checking in `jackdaw.specs`. We probably want `jackdaw.streams` to use them.

PR #102, opened 2019-03-01
Word Count Modernization

creese:
This updates the topic config to use the Jackdaw Serdes resolver and moves the core topology out of the comment block so we can add tests.

Other notable changes:
- Removes the Confluent namespace from the path.
- Expands doc-strings and the documentation in the remaining comment block.
- Adds logging.
- Suppress warn level log messages having to do with unknown config `admin.retries`.
- Updates the README.
- Updates `jackdaw.streams.mock` style test.
- Modifies the EDN serde so it can be used from Java and sets this as the default serde for Word Count


examples/word-count/src/word_count.clj
```diff
@@ -13,34 +9,33 @@
...
 
+   "materialized"
+   {:key-serde {:serde-keyword :jackdaw.serdes.edn/serde}
+    :value-serde {:serde-keyword :jackdaw.serdes.edn/serde}}})
 
-;;; App Template
-;;;
+(def topic-metadata
+  (memoize (fn []
+             (fmap #(assoc % :key-serde ((resolver/serde-resolver) (:key-serde %))
```
cddr: Shall we just use a single `serde-resolver`?

examples/word-count/src/word_count.clj
```diff
@@ -13,34 +9,33 @@
...
     :replication-factor 1
-    :key-serde key-serde
-    :value-serde value-serde}))
+    :key-serde {:serde-keyword :jackdaw.serdes.edn/serde}
+    :value-serde {:serde-keyword :jackdaw.serdes.edn/serde}}
 
-(defn topic-names
-  []
-  ["input" "output"])
+   "output"
```
cddr: Would be cool if these were keywords. Then it could be re-used by the test-machine code.

PR #103, opened 2019-03-06
Changed topic->produced to consider partition-fn as well

funderbar:
This PR enables the ability to use `Produced/With` function with the arity that has `StreamPartitioner`, so that we can use through with the partition function.

Basically new through arity will accept the `Produced` interface using the following function.
```java
public static <K, V> Produced<K, V> with(final Serde<K> keySerde,
                                         final Serde<V> valueSerde,
                                         final StreamPartitioner<? super K, ? super V> partitioner) {
     return new Produced<>(keySerde, valueSerde, partitioner);
}
```


src/jackdaw/streams/specs.clj
```diff
@@ -49,7 +49,8 @@
 (s/def ::topic-config
   (s/keys :req-un [::topic-name
                    ::key-serde
-                   ::value-serde]))
+                   ::value-serde]
+          :opt-un [::partition-fn]))
```
creese: Can we add this to the specs in `jackdaw.specs` too?

test/jackdaw/streams_test.clj
```diff
@@ -198,20 +198,38 @@
...
+                                        (-> builder
+                                            (k/kstream topic-a)
+                                            (k/through topic-b)
+                                            (k/to topic-c))))
+            publish (partial mock/publish driver topic-a)]
+
+        (publish 1 1)
+
+        (is (= [[1 1 10]] (mock/get-keyvals driver topic-b)))
+        (is (= [[1 1]] (mock/get-keyvals driver topic-c))))))
```
creese: `jackdaw.streams.mock/get-keyvals` might be doing the wrong thing here. When more than a key and value is returned, the result should probably be a map. See `jackdaw.client/poll` for the expected behavior.

funderbar: This was an intermediary push. Can you check its current state

src/jackdaw/streams/specs.clj
funderbar: will do

src/jackdaw/specs.clj
```diff
@@ -1,15 +1,21 @@
...
 (s/def ::topic-name
   string?)
 
+(s/def ::partition-fn
+  ct/function?)
+
 (s/def :jackdaw.topic/topic
-  (s/keys :req-un [::topic-name]))
+  (s/keys :req-un [::topic-name]
+          :opt-un [::partition-fn]))
```
creese: Please add `::partition-fn` to `:jackdaw.serde-client/topic` as well.

src/jackdaw/specs.clj
```diff
@@ -1,15 +1,21 @@
...
+
 
 
 ;; The basic topic
 
 (s/def ::topic-name
   string?)
 
+(s/def ::partition-fn
+  ct/function?)
```
creese: Let's try `ifn?`.

src/jackdaw/specs.clj
```diff
@@ -45,7 +50,8 @@
   (s/keys :req-un [::topic-name
                    ::partition-count
                    ::replication-factor
-                   ::topic-config]))
+                   ::topic-config]
+          :opt-un [::partition-fn]))
```
creese: `::partition-fn` isn't needed to create a topic. We should remove this.

src/jackdaw/streams/mock.clj
```diff
@@ -80,24 +81,37 @@
   ([test-driver topic-config time-ms k v]
    ((producer test-driver topic-config) time-ms k v)))
 
+; XXX(Ozum): rename to key-val-extractor ?
 (defn producer-record
```
creese: Is this still needed?

test/jackdaw/streams_test.clj
```diff
@@ -333,8 +357,8 @@
 
       (produce-a 1 1 1)
       (produce-b 100 2 2)
-      (is (= [1 1] (mock/consume driver topic-c)))
-      (is (= [2 2] (mock/consume driver topic-c)))))
+      (is (= [1 1] (mock/get-keyval driver topic-c)))
```
funderbar: get-keyval was particularly made for here.
We could easily call consume directly and convert to key-val as well

PR #111, opened 2019-03-12
Imported jackdaw.serdes namespace to resolver

funderbar:
When we attempt to use the jackdaw resolver with the keyword
`:jackdaw.serdes/string-serde`, the keyword cannot be resolved into the 
serde function it is supposed to be. So we import the namespace into the resolver.

### DIFF
--- a/src/jackdaw/serdes/resolver.clj
+++ b/src/jackdaw/serdes/resolver.clj
@@ -5,6 +5,7 @@
             [jackdaw.serdes.avro.confluent]
             [jackdaw.serdes.edn]
             [jackdaw.serdes.json]
+            [jackdaw.serdes]
             [jackdaw.specs]))
 
 (defn load-schema


src/jackdaw/serdes/resolver.clj
```diff
@@ -5,6 +5,7 @@
             [jackdaw.serdes.avro.confluent]
             [jackdaw.serdes.edn]
             [jackdaw.serdes.json]
+            [jackdaw.serdes]
```
cddr: The jackdaw.test.transport ns resolves a similar problem by using a macro in order to expose which multi-methods have been defined so that it can be tested. Maybe overkill but nice to have in-case someone runs clj-refactor on the namespace and tries to remove it.

https://github.com/FundingCircle/jackdaw/blob/90868724b77bfd03909a5fb2a06da7f3025a4303/src/jackdaw/test/transports.clj#L3

PR #112, opened 2019-03-12
Created a new string namespace in jackdaw.serdes

funderbar:


cddr:
I think having lots of different namespaces all with function of the same name (in this case `serde`) is a bit of an anti-pattern and if we're going to re-organize this namespace, I'd prefer if we went in the other direction. To illustrate, consider the case where you want to use a bunch of different serdes in some program....

This would be my preference as a *user* of jackdaw...

```clojure
(ns my.cool.prog
  (:require
     [jackdaw.serdes :as serde]))

(def topics
  {:foo {:key-serde (serde/string-serde)
            :value-serde (serde/avro-serde avro-options)}
    :bar {:key-serde (serde/integer-serde)
             :value-serde (serde/edn-serde)}
    :baz {:key-serde (serde/long-serde)
             :value-serde (serde/json-serde)})
```

vs

```clojure
(ns my.cool.prog
  (:require
     [jackdaw.serdes.edn :as edn-serde]
     [jackdaw.serdes.string :as string-serde]
     [jackdaw.serdes.json :as json-serde]
     [jackdaw.serdes.avro :as avro-serde]))

(def topics
  {:foo {:key-serde (string-serde/serde)
            :value-serde (avro-serde/serde avro-options)}
    :bar {:key-serde (integer-serde/serde)
             :value-serde (edn-serde/serde)}
    :baz {:key-serde (long-serde/serde)
             :value-serde (json-serde/serde)})
```

99-not-out:
Stale PR

PR #119, opened 2019-03-15
Use lein-tool-deps

AndreaCrotti:
Use this project to allow a smoother transition to deps.edn
https://github.com/RickMoynihan/lein-tools-deps

Only dependencies information are moved into deps.edn, all the rest stays in project.clj

### DIFF
--- a/deps.edn
+++ b/deps.edn
@@ -1,0 +1,13 @@
+{:mvn/repos {"confluent" {:url "https://packages.confluent.io/maven/"}}
+ :deps {aleph {:mvn/version "0.4.6"}
+        clj-time {:mvn/version "0.15.1"}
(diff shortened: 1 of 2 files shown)


gphilipp:
Actually, I advise against using the `lein-tools-deps` plugin because of the flattening issue: https://github.com/RickMoynihan/lein-tools-deps/issues/30#issuecomment-388331765

cddr:
Going to close this. If we really want to move to deps, I think we can go pure deps and just use maven to publish/deploy.

RickMoynihan:
FYI: there is a potential solution to the flattening issue described here:

https://github.com/RickMoynihan/lein-tools-deps/issues/91

Or if deploys are all that is holding you back from adopting tools deps, then you may consider using https://github.com/slipset/deps-deploy

PR #121, opened 2019-03-18
Fix typo in service-ready? fixture

cddr:
This fixes a typo in the 2nd condition of the `service-ready?` fixture which could cause the service to be declared ready before it is actually responding with an HTTP code of 200.

This might be what is causing the rest-proxy test to sometimes fail due to commencing the test before the service is properly ready.


creese:
This is a big changeset. Did you mean to use `master` for the base?

PR #126, opened 2019-04-18
Make avro serde accept clj maps with underscores, add option to skip mangling names.

sathyavijayan:
Ive made a quick fix/change to the avro serde to make it accept non-idiomatic, but perfectly valid maps as input. For eg.
```
{"string_field": "hello", :enum_field "a_1"}
```
I also added an option to skip mangling names like in the [abracad library](https://github.com/damballa/abracad/blob/master/src/clojure/abracad/avro/util.clj#L42). This will be useful when we need to use the original message.

### DIFF
--- a/src/jackdaw/client.clj
+++ b/src/jackdaw/client.clj
@@ -281,10 +281,10 @@
   [^Consumer consumer timestamp topics]
   (let [topic-partitions (->> (mapcat #(partitions-for consumer %) topics)
                               (map #(select-keys % [:topic-name :partition])))
+        end-offsets      (end-offsets consumer topic-partitions)
         ts-offsets       (offsets-for-times consumer
                                          (zipmap topic-partitions
                                                  (repeat (count topic-partitions) timestamp)))
-        end-offsets      (end-offsets consumer topic-partitions)
         offsets          (reduce-kv (fn [m k v]
                                       (assoc m k {:ts-offset v
                                                   :end-offset (get end-offsets k)}))


test/jackdaw/serdes/avro_test.clj
```diff
@@ -457,6 +457,199 @@
...
+                          :type ["null" {:name "subrecords"
+                                         :type "array"
+                                         :items "banana"}]}
+                         {:name "uuid_field"
+                          :type {:type "string",
+                                 :logicalType "uuid"}}]}
+        schema-str (json/write-str schema)
+        serde      (->serde schema-str)]
+
+    (is (= (round-trip serde "bananas"
```
cddr: It took me a while before I realized that the difference between the 1st and 2nd round-trip tests here was that the enum field was expressed as a string in the first, and a keyword in the second. Would you mind putting each one inside a `testing` block so that they can more easily be distinguished?

cddr (commented):
I think this would be a useful change. If I had our time again, `*mangle-names* == false` would be the default but I think the ship has sailed on that.

@creese do you accept this in principle? If so, we might consider adding it as an option here https://github.com/FundingCircle/jackdaw/blob/630768902fdfac602f91b809add0a63411dd44de/src/jackdaw/serdes/avro.clj#L607 so that end-users don't need to fiddle with dynamic vars.

test/jackdaw/serdes/avro_test.clj
sathyavijayan: Thanks. I have restructured the tests and added documentation to make them more readable. 

I used a dynamic var for `*mangle-names* ` as it made more sense for me to have it as an 'application-level configuration'. It will lead to a consistent low level design within an application  - either use idiomised values or not. (although, I agree the degree of ambiguity wouldn't be much as there will always be a default setting). 

I noticed that the mangling is done only for the serde and not for json etc., So, as you mentioned, it would make sense to drop it altogether and push the responsibility to the apps if they need such a feature. The dynamic var will be a nice way to provide backward compatibility during transition if you decide to drop mangling in a future release.

Let me know your thoughts.

sathyavijayan: Hi @cddr, @creese any thoughts on this PR ? Let me know if you still like the dyn var to be moved to the `topic-config` map (last arg to the serde function) instead. The map is named `topic-config` which makes it a weird place for the `mangle-names` key - what do you think ?

peey:
Anything blocking this from being merged?

obohrer:
Can this be merged? This caused us issues as well as the data mangling was a surprise to us

PR #135, opened 2019-05-10
Coerce clojure numbers to avro numbers

matthias-margush:
Coerces clojure number values to avro int/long when possible.


kisakov:
Can we please add these changes into next release?

PR #139, opened 2019-06-03
Adjust test partition fn arity to what is expected from kafka streams

funderbar:


src/jackdaw/test/commands/write.clj
```diff
@@ -4,7 +4,7 @@
    [clojure.tools.logging :as log]
    [jackdaw.client.partitioning :as partitioning]))
 
-(defn default-partition-fn [topic-map k]
+(defn default-partition-fn [topic-name topic-map k v]
```
creese: `topic-name` is a key in `topic-map`. I think we should match the signature here:
https://github.com/FundingCircle/jackdaw/blob/0ae1d7631816572606334e222f16412d9dc56d6b/test/jackdaw/streams_test.clj#L218

funderbar: You are definitely right Charles, not sure how I missed this, I think I was trying to make `default-partition-fn` usable and I made the other one unusable. Check the current change and let me know what you feel.

PR #141, opened 2019-06-07
Add more test machine commands for inspecting its state

99-not-out:
Added two new Test Machine commands.

`:do!` - like `do` but the function provided is passed the journal atom, not the contents of the journal atom.

`:inspect` - the function provided is passed the whole test machine state (a debugging hook)

### DIFF
--- a/doc/test-machine.md
+++ b/doc/test-machine.md
@@ -100,7 +100,16 @@ specific arguments. Currently the following commands are the supported.
   :sleep    [sleep-ms]           Sleeps for `sleep-ms` milliseconds
   :println  [args]               Prints the supplied args to stdout
   :pprint   [args]               Pretty prints the supplied args to stdout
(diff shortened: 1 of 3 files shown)


cddr:
Seems reasonable. Can you add the new commands to the docs plz?

https://github.com/FundingCircle/jackdaw/blob/master/doc/test-machine.md#test-commands

creese:
Seems like one is a superset of the other? Should we just change how `:do` behaves?

99-not-out:
Well, the difference between `do` and `do!` is that do will give you a static snapshot of the journal state at the time `do` ran. Where as `do!` returns the atom, which is therefore still being updated in the background. So `do!` lets you inject some custom journal update monitoring or whatever code.

`do!` and `inspect` are really debugging commands as a result.

PR #144, opened 2019-06-13
Stop swallowing errors in test-machine

cddr:
When working with the test-machine, it makes for a significantly better developer workflow when any exceptions that occur are bubbled back up to the main thread. Prior versions were catching errors and sometimes not even logging them (mea culpa!).

This PR adds tests with a few of the errors we've seen developers make and ensures that the resulting exception (including the full cause trace) is logged and the exception is bubbled back up to the main thread after all the resources used by the test are torn back down.

### DIFF
--- a/src/jackdaw/test/journal.clj
+++ b/src/jackdaw/test/journal.clj
@@ -20,10 +20,8 @@
         check-condition (fn [new]
                           (when-let [result (condition? new)]
                             (remove-watch journal id)
-                            (if (boolean? result)
-                              (deliver p {:result :found})
-                              (deliver p result))))]
-
+                            (deliver p {:result :found
+                                        :info result})))]
     (add-watch journal id (fn [k r old new]
                             (check-condition new)))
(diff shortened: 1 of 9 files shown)


src/jackdaw/test/journal.clj
```diff
@@ -20,10 +20,8 @@
         check-condition (fn [new]
                           (when-let [result (condition? new)]
                             (remove-watch journal id)
-                            (if (boolean? result)
-                              (deliver p {:result :found})
-                              (deliver p result))))]
-
+                            (deliver p {:result :found
```
cddr: This change addresses another usability problem encountered in the wild. Test authors could confuse the middleware if they wrote a `condition?` function that returned a map containing a `:result` key. The original intention was that the return value might be of interest to the test author when making assertions about the results but clearly it is safer to include the result in it's own `:info` key to make sure it doesn't collide with what the middleware is expecting to see.

PR #143, opened 2019-06-12
Fix kstreams

sbrauer:
Before this change, trying to use `kstreams` would fail because it was
missing the Consumed argument.
The newly added test would throw an exception like this:

```
Error in KStream
Uncaught exception, not in assertion
   error: org.apache.kafka.streams.errors.StreamsException: Exception caught in process. taskId=0_0, processor=KSTREAM-SOURCE-0000000000, topic=topic-a, partition=0, offset=0, stacktrace=org.apache.kafka.streams.errors.StreamsException: A serializer (key: org.apache.kafka.common.serialization.LongSerializer / value: org.apache.kafka.common.serialization.LongSerializer) is not compatible to the actual key or value type (key type: clojure.lang.Symbol / value type: clojure.lang.Symbol). Change the default Serdes in StreamConfig or provide correct Serdes via method parameters.
    at org.apache.kafka.streams.processor.internals.SinkNode.process(SinkNode.java:94)
    at org.apache.kafka.streams.processor.internals.ProcessorContextImpl.forward(ProcessorContextImpl.java:183)
    at org.apache.kafka.streams.processor.internals.ProcessorContextImpl.forward(ProcessorContextImpl.java:162)
    at org.apache.kafka.streams.processor.internals.ProcessorContextImpl.forward(ProcessorContextImpl.java:122)
    at org.apache.kafka.streams.processor.internals.SourceNode.process(SourceNode.java:87)
    at org.apache.kafka.streams.processor.internals.StreamTask.process(StreamTask.java:364)
    at org.apache.kafka.streams.TopologyTestDriver.pipeInput(TopologyTestDriver.java:406)
    at sun.reflect.GeneratedMethodAccessor10.invoke(Unknown Source)
    at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
    at java.lang.reflect.Method.invoke(Method.java:498)
    at clojure.lang.Reflector.invokeMatchingMethod(Reflector.java:93)
    at clojure.lang.Reflector.invokeInstanceMethod(Reflector.java:28)
    at jackdaw.streams.mock$producer$produce_BANG___8206.invoke(mock.clj:74)
    at jackdaw.streams.mock$publish.invokeStatic(mock.clj:80)
    at jackdaw.streams.mock$publish.invoke(mock.clj:78)
    at clojure.core$partial$fn__5561.invoke(core.clj:2618)
[...]

PR #149, opened 2019-06-19
Select matching record from union for coercion during deserialisation.

kelveden:
During the coercion phase of avro deserialisation, just assuming that a given `GenericData$Record` is the correct one purely on the basis that it is a `GenericData$Record` isn't enough if the record in question is one of several in a union type - the result is that the first record from the union every time.

This PR changes the logic so that the given `Record` only is considered to "match" if the schema itself matches the schema attached to the `Record` by avro itself.

### DIFF
--- a/src/jackdaw/serdes/avro.clj
+++ b/src/jackdaw/serdes/avro.clj
@@ -390,8 +390,12 @@
               field->schema+coercion)))
 
   (match-avro? [_ avro-record]
-    (or (instance? GenericData$Record avro-record)
-        (nil? avro-record)))
+    (cond
+      (nil? avro-record) true
+
+      (instance? GenericData$Record avro-record)
+      (let [^GenericData$Record generic-data-record avro-record]
+        (= schema (.getSchema generic-data-record)))))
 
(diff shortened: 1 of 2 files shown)


cddr:
Ah one thing I forgot. Please can you update the commit message to include a line like

"Signed-off-by: First Last <email>"

A good example is this one.

https://github.com/FundingCircle/jackdaw/pull/143/commits

kelveden:
Done @cddr. Maybe worth an update in the "Contributing" section of the README?

kelveden:
Not sure what to do exactly about that `deploy_snapshot` failure in the build.

cddr:
> Not sure what to do exactly about that deploy_snapshot failure in the build.

It's ok for that to fail. That's just because you don't have our key for deploying snapshots to clojars.

> Maybe worth an update in the "Contributing" section of the README?

Yep. This policy has only recently been put in place and most of the contributors until now have been internal. I updated https://github.com/FundingCircle/jackdaw/blob/master/CONTRIBUTING.md just after commenting on your PR but I'll add a link to that doc from the README too. Thanks again.

PR #150, opened 2019-06-19
Add docs to minimize git sign-off overhead

cddr:
Adds documentation so that it is easy for contributors to set their local environment up to do the right thing vis-a-viz git sign-off but also ensure it's just a one-liner to fix up the history when we forget.

Signed-off-by: Andy Chambers <andy.chambers@fundingcircle.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,20 @@ of the manual][8].  Adding information like the backtrace and any sample message
 and/or topic configurations used to the bug report makes it easier to track down bugs.
 Some steps to reproduce a bug reliably would also make a huge difference.
 
+## Signoff on Commits
+
+```
+# signoff an individual commit
+git commit -s -m 'Add foo feature'
+
+# signoff all commits by default
+git config format.signoff true
+
+# git alias to rebase sequence of commits to include signoff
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -30,6 +30,20 @@ of the manual][8].  Adding information like the backtrace and any sample message
...
+## Sign-off on Commits
+
+```
+# sign-off an individual commit
+git commit -s -m 'Add foo feature'
+
+# sign-off all commits by default
+git config format.signoff true
+
+# git alias to rebase sequence of commits to include sig-noff
```
creese: I think you have a typo, `sig-noff`, should be `sign-off`.

CONTRIBUTING.md
```diff
@@ -30,6 +30,20 @@ of the manual][8].  Adding information like the backtrace and any sample message
 and/or topic configurations used to the bug report makes it easier to track down bugs.
 Some steps to reproduce a bug reliably would also make a huge difference.
 
+## Sign-off on Commits
```
creese: In this context, I think "sign off" is a verb. There is no hyphen. The same is probably true for some of the comments below.

PR #154, opened 2019-06-20
delete duplicated test

AndreaCrotti:
`test-watch-command-error` is defined in the same way just above.


AndreaCrotti:
I'll do it again from our own fork so no external contributors needed.

PR #155, opened 2019-06-20
use a macro for with-test-machine

AndreaCrotti:


AndreaCrotti:
Uhm the fact that `test-a` and `test-b` take a machine in input and return a machine it's a bit weird imho.
And I also thought that since the config is just a vector, all the composition would be done that way.

Anyway this was just an experiment to see if it worked and I opened a PR just to see what CircleCI had to say.
I need to have a deeper look to see if it's worth, and if we don't want to change it in any case for backward compatibility well we can keep it this way.
(We could still add an extra macro though right without changing the existing one right?)

cddr:
> Uhm the fact that test-a and test-b take a machine in input and return a machine it's a bit weird imho.

Slightly non-idiomatic for sure. But I'm too old to worry about what people think is weird.

Don't get me wrong. I love a [good macro](https://github.com/cddr/celtk/blob/bcf52025dc32c14d20dd37ff186d9f18dc6266b7/tk-object.lisp#L54). But I think the job of this library is to provide solid functions that operate on well-defined data structures so that application developers can build their own macros and/or higher-order functions tailored to their own use-case.

99-not-out:
It would be good to start some form of high-level wrapper code for TestMachine, as there are a variety of good helpers we already have over things like:
* Building instructions
* Getting stuff out of the journal
* Running the machine
...

These should be layered & composed up over the core, and would allow experimentation like this in a sandbox for now, and promotion of the nice bits as we go.

AndreaCrotti:
Closing this in the meanwhile, can give it another go later on

Issue #152, opened 2019-06-20
Expose "reader schema" feature


cddr:
Avro has a very interesting ability to use a different schema for **reading** a message than the one used to **writing** the message. Both schemas must be made available to the reader (this is the _raison d'être_ for the schema registry) but as long as the schemas are compatible, it means that consumers and producers are not forced to share their schemas via some centralized github repo.

At the moment jackdaw doesn't support this feature. Even though it requires a `:value-serde` in many consuming contexts, it does not actually **use** it to deserialize the message.

It should be possible to add a `:reader-schema` key to avro serde multi-method which allows users to take advantage of this feature without breaking any existing uses.

cddr:
Resolved by #157

Issue #137, opened 2019-05-17
Avro serde requires a local schema even for Consumers


cddr:
The avro serde resolver requires the specification of a local schema even though the schema is only actually used for *producing* messages. When consuming messages, we just let the `KafkaAvroDeserializer` lookup the schema-id in the registry so when "resolving" a topic-metadata definition we should not require the presence of a schema.

PR #157, opened 2019-06-25
Decouple avro reader and writer serdes

### LINKED ISSUE #136: Schema registry basic auth issues
I'm not sure if its how jackdaw creates its default client or not but I couldn't find a way to set or get it to pick up the piece of config to tell it to extract the basic auth creds from the URL. A simple workaround is to create one yourself

```
(defn cached-schema-registry-cli []
  (let [rest-service (RestService. "https://user:pass@schema-reg.com:13584")]
    (CachedSchemaRegistryClient. rest-service 10 (doto (new java.util.HashMap) (.put "basic.auth.credentials.source" "URL")))))

(def serde-resolver
  (partial resolver/serde-resolver :schema-registry-url "" :schema-registry-client (cached-schema-registry-cli)))
```

Minor points but it would also be nice to not require a url for the resolver and also the topic-metadata map insists on a default schema even if using it from the registry.

Thanks.

cddr:
I believe this resolves a few issues including #152, #137. It also represents part of the work required to resolve #136 although that might need a bit more work on the serialization side and could certainly use a test and documentation to demonstrate it's use. I plan to follow up on this with another guide in the docs section all about configuring the various kafka client entities (e.g. producers, consumers, serdes etc).

The main change is to add a `:deserialization-properties` optional keyword to the confluent serde constructor. Including the property "specific.avro.reader" in these properties means that any schema supplied to the deserializer will be used in combination with the writer's schema to decode the message. The reader schema must be compatible with the writer's schema in order for this to work correctly. See the schema resolution rules in the related links for the details about how this works.

Another change is that it is no longer an error if you try to create an avro serde without specifying a local copy of the schema. This can be used to consume data produced by some other process that *does* have access to the schema or which generates it's own schema on the fly (e.g. kafka-connect, debezium etc).

Finally I tried to remove some of the magic from the serde-resolver function. It should remain logically the same (except for not throwing an error as described above) but it is hopefully a bit clearer what's happening now.

Related Links:

 * https://avro.apache.org/docs/1.8.1/spec.html#Schema+Resolution

### DIFF
--- a/src/jackdaw/serdes/avro.clj
+++ b/src/jackdaw/serdes/avro.clj
@@ -516,15 +516,19 @@
     clj-serializer))
 
 (defn- deserializer [schema->coercion serde-config]
-  (let [{:keys [registry-client registry-url avro-schema key?]} serde-config
-        base-deserializer (KafkaAvroDeserializer. registry-client)
+  (let [{:keys [registry-client registry-url avro-schema key?
+                deserializer-properties]} serde-config
+        base-deserializer (KafkaAvroDeserializer. registry-client (let [min-props {"schema.registry.url" registry-url}]
+                                                                    (merge min-props deserializer-properties)))
         methods {:close       (fn [_]
                                 (.close base-deserializer))
                  :configure   (fn [_ base-config key?]
                                 (.configure base-deserializer base-config key?))
                  :deserialize (fn [_ topic raw-data]
                                 (try
-                                  (let [avro-data (.deserialize base-deserializer topic raw-data)]
+                                  (let [avro-data (if (get deserializer-properties "specific.avro.reader")
+                                                    (.deserialize base-deserializer topic raw-data avro-schema)
+                                                    (.deserialize base-deserializer topic raw-data))]
                                     ;; Note that `.deserialize` will return EITHER a Java Object, or
                                     ;; a ^GenericContainer. ^GenericContainer is only produced when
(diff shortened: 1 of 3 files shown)


src/jackdaw/serdes/avro.clj
```diff
@@ -600,7 +606,8 @@
     :as   registry-config}
    {:keys [avro/schema
            avro/coercion-cache
-           key?]
+           key?
+           deserializer-properties]
```
creese: I think this is trying to do too many things at once. If we want a SerDes that takes a reader schema, we should probably pass it two args with the writer schema required and the reader schema optional. If we just want a deserializer that returns a SerDes that can only be used to consume, I think this should be a separate top level function. (Without a writer schema, it no longer returns a "serde".) How about `deser`?

99-not-out:
I would perhaps go further here open it up more - if we allow the `serailizer` and `deserializer` fns here https://github.com/FundingCircle/jackdaw/blob/7e4e5e7be717334b75b04242deb5971f6d52b479/src/jackdaw/serdes/avro.clj#L642 to be optionally passed in then the user can experiment with stuff like this all they want.

This allows overriding the default behaviour completely, and we can provide several versions of these two fns if we find good candiates - or eventually alter the default behaviour.

But such a change would allow experimentation in this area without so much local overriding (which is hard as the two fns in question are currently private).

PR #159, opened 2019-06-26
Upgrade clojure to 1.10.1

MicahElliott:
Jackdaw already requires its users to use 1.10, so it should use it
too.  Aviso logging needed a bump up to avoid a syntax error in
previous 0.3.1.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -22,7 +22,7 @@
                  [org.apache.kafka/kafka-streams "2.2.0"]
                  [org.apache.kafka/kafka_2.11 "2.2.0"]
                  [org.apache.kafka/kafka-streams-test-utils "2.2.0"]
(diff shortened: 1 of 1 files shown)


kisakov:
> Jackdaw already requires its users to use 1.10, so it should use it
I'm asking mostly for curiosity.
@MicahElliott  Do you know where Jackdaw requires to use 1.10? Or is there any errors if will use new jackdaw with 1.9 in my project?

cddr:
@kisakov All the stuff in [here](https://github.com/FundingCircle/jackdaw/tree/master/src/jackdaw/data) uses the datafy protocol which was introduced in 1.10. There was an attempt to make it work without but it didn't work as intended.

PR #160, opened 2019-06-27
Add doc/article for the jackdaw client API

cddr:
Documentation (in the form of an article) for the lower-level client API. Includes simple examples for both consuming and producing data. Tries to highlight the key points required to understand how to use the API and provide links to the relevant upstream documentation where applicable.

### DIFF
--- a/doc/clients.md
+++ b/doc/clients.md
@@ -6,53 +6,77 @@ The Jackdaw Client API wraps the core Kafka `Producer`<sup>[1](#producerapi)</su
 `Consumer`<sup>[2](#consumerapi)</sup> APIs and provides functions for building or
 unpacking some of the supporting objects like Callbacks, Serdes, ConsumerRecords etc.
 
+While Kafka's surface API is quite small, the functionality it provides is deep. You
+can get up and running very quickly with a simple example but to fully understand it's
+capabilities there is no substitute for reading the upstream documentation. The
+scope of this guide therefore is limited to demonstrating how to use API via Jackdaw
+and connecting the relevant parts of the documentation for further reading.
+
 ## Producing
 
 The producer example below demonstrates how to use the Kafka Producer API. The configuration<sup>[3](#producerconfig)</sup>
-is represented as a simple map (Jackdaw will convert this to a `Properties` object), and the
-key and value serializers are specified when creating the producer to override the default
-(which would be the `StringSerializer`).
+is represented as a simple map (Jackdaw will convert this to a `Properties` object) and in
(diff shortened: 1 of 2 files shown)


doc/clients.md
```diff
@@ -0,0 +1,96 @@
+# Jackdaw Client API
```
creese: Should the filename be `doc/client.md` without the 's'?

doc/clients.md
```diff
@@ -0,0 +1,96 @@
...
+(ns producer-example
+  (:require
+    [jackdaw.serdes.json :as json]
+    [jackdaw.client :as jc])
+  (:import
+    (org.apache.kafka.common.serialization Serdes)))
+
+(def producer-config
+  {"bootstrap.servers" "localhost:9092"
+   "acks" "all"
```
creese: Is this required or just good practice?

doc/clients.md
```diff
@@ -0,0 +1,96 @@
...
+and offset of the written record.
+
+In this example, the JSON serde converts the message from a plain old Clojure map into a
+JSON byte-array while building the ProducerRecord that is eventually passed to the
+KafkaProducer's `.send` method. Other Serdes are available<sup>[4](#serdesdirectory)</sup>
+
+```
+(ns producer-example
+  (:require
+    [jackdaw.serdes.json :as json]
```
creese: If we use a string serializer for the value, we can remove this dependency.

doc/clients.md
cddr: The thinking behind the plural was that this covered both types of clients, producers and consumers. I can appreciate the argument for making it match the final segment of the namespace. Not too fussed either way.

doc/clients.md
cddr: Yeah that's probably a good idea. I think it will be good to describe somewhere how serdes can be used to automatically marshall plain old clojure data into whatever serialization format is desired but that probably merits its own doc.

doc/clients.md
cddr: Ah yeah I think I was going to add a note about that describing the other options available and the trade-offs. I'll add that and ♻️.

doc/clients.md
creese: Two comments:
1. Kafka treats the consumer and producer as separate APIs, but in Jackdaw we treat these as a single "Client API". 
2. I was thinking we may want to match the Clojure namespace the documentation is for.

This is not a blocker.

cddr: Agreed. I've updated to client.md

PR #172, opened 2019-07-24
Supply `key-serde` as well as `value-serde` in `aggregate` methods

lxsli:
There's already a `topic->materialized` helper function, which I've
reused.

Signed-off-by: Alexis Lee <alee@greshamtech.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [0.6.7] - [2019-07-24]
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


lxsli:
This doesn't seem to work... I'm wondering if Kafka also ignores it, for some reason

lxsli:
Welp, new tests pass so I guess Jackdaw is fine now, the bug I'm seeing in my private code must be something else. PR should be good to merge.

99-not-out:
@cddr I will take a look - thanks!

PR #174, opened 2019-07-26
New feature: Add schema metadata to deserialized avro records

sbrauer:
~~When an Avro record is deserialized to a Clojure map, add the record's name and fully-qualified name to the map as metadata.
These record names are potentially useful info that was being discarded.~~

When an Avro record is deserialized to a Clojure map, add the record's schema to the map as metadata.

My personal use case for this metadata is to make it easier for consuming code to deal with unions of records.

### DIFF
--- a/src/jackdaw/serdes/avro.clj
+++ b/src/jackdaw/serdes/avro.clj
@@ -399,21 +399,24 @@
 
   (avro->clj [_ avro-record]
     (when avro-record
-      (into {}
-            (comp (map first)
-                  (map (fn [^Schema$Field field]
-                         (let [field-name (.name field)
-                               field-key (keyword (unmangle field-name))
-                               [_ field-coercion :as entry] (get field->schema+coercion field-key)
-                               value (.get ^GenericData$Record avro-record field-name)]
-                           (when-not field-coercion
-                             (throw (ex-info "Unable to deserialize field"
-                                             {:field field
-                                              :field-name field-name
-                                              :field-key field-key
-                                              :entry entry})))
-                           [field-key (avro->clj field-coercion value)]))))
-            (vals field->schema+coercion))))
+      (with-meta
+        (into {}
+              (comp (map first)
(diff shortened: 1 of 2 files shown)


src/jackdaw/serdes/avro.clj
```diff
@@ -399,21 +399,24 @@
...
+                             (when-not field-coercion
+                               (throw (ex-info "Unable to deserialize field"
+                                               {:field field
+                                                :field-name field-name
+                                                :field-key field-key
+                                                :entry entry})))
+                             [field-key (avro->clj field-coercion value)]))))
+              (vals field->schema+coercion))
+        {:name (.getName schema)
+         :fullname (.getFullName schema)})))
```
cddr: I think when hyphenating the `FullName` symbol I'd expect to see `full-name`.

sbrauer: I just pushed a commit with the hyphenated name.

sbrauer:
I hope you agree that there's value in exposing the record names. There may be other ways to do that besides metadata (although it was the best way I could think of), and even with the metadata approach one might have chosen different keywords.
I'd be happy to discuss.

cddr:
Hey Sam. Sorry for the radio silence so far. The other person in our company who might have an opinion on this is away on holiday at the moment.

I agree that this is valuable information that should be made available and that given the current setup, the metadata seems like the only place to put it. I wonder whether we could just put the whole schema in the metadata (or maybe even both a writer schema and a reader schema).

sbrauer:
No worries. Good to hear that you're open to the general idea. I've also been thinking that exposing the schema itself could be better than exposing the names (since the names and a lot of other info can be obtained from the schema; on the other hand one could argue it's exposing too much internal detail).

I must admit that I haven't yet taken the time to digest your recent change related to decoupling the reader and writer schemas. I should do that soon. In the meantime I pushed a commit that adds the schema to the metadata instead of the names. I'll update the PR title as well.

PR #175, opened 2019-07-31
Allow inner joining a kstream to a ktable.

GreshamDanielStephens:
Add support for inner joins between a kstream and ktable.
This is supported by the java api but not yet exposed in jackdaw.
Moves the existing `join` method from KTable to KStreamBase to be shared.

Signed-off-by: Daniel Stephens <dstephens@greshamtech.com>


zerg000000:
@GreshamDanielStephens any progress on this?

GreshamDanielStephens:
Ahh that's a shame, I will try and get some more tests written this weekend.

Issue #176, opened 2019-08-02
`reset-application-fixture` hardcodes bootstrap.servers value


timclemons:
Attempting to use `reset-application-fixture` or, by extension, `integration-fixture` with a Kafka bootstrap server other than localhost:9092 results in a connection timeout.  Examining the source, it looks like this value has been hardcoded as an argument to `StreamsResetter`.

cddr:
Thanks for the bug report @timclemons. @7vs is taking a look at this

cddr:
Closed by #177

PR #177, opened 2019-08-09
[ISSUE-176] allow reset-application-fixture to get bootsrap.servers from app-config

7vs:
Issue #176 was already fixed by #179 although the fix was not backed by a unit test. @7vs and I (@cddr) had already begun working on this when @kidpollo submitted #179 but figured we'd continue as this PR back-fills a unit-test for the feature and also adds the ability for the user of the fixture to supply optional parameters, and an optional `reset-fn` which allows the unit test to check that the error case is handled with an informative exception (the StreamsResetter quite often fails in a dev environment when the state can easily become inconsistent but the error message usually contains enough information about what to do to resolve the situation.

### DIFF
--- a/src/jackdaw/test/fixtures.clj
+++ b/src/jackdaw/test/fixtures.clj
@@ -197,6 +197,8 @@
 ;;; reset-application-fixture ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (defn reset-application-fixture [app-config reset-fn reset-args]
+  "Reset application with default app-config as well as reset-args,
+   but the fixture can implement custom options"
   (fn [t]
     (let [rt (StreamsResetter.)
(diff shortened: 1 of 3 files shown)


src/jackdaw/test/fixtures.clj
```diff
@@ -201,7 +201,7 @@
     (let [rt (StreamsResetter.)
           app-id (get app-config "application.id")
           args (->> ["--application-id" (get app-config "application.id")
-                     "--bootstrap-servers" "localhost:9092"]
+                     "--bootstrap-servers" (or (get app-config "bootstrap-servers") "localhost:9092")]
```
cddr: Think "bootstrap-servers" here should be "bootstrap.servers" but LGTM apart from that.

7vs: oh, sorry, I forgot that little detail :) you are correct

PR #184, opened 2019-08-23
Better test runner

AndreaCrotti:
- use kaocha to run tests (https://github.com/lambdaisland/kaocha)
- merge the only test profile setting into dev
- remove lein-cloverage
- support the test failure summary in circleci by publishing the test results

A failure will look like this with this change:

Screenshot 2019-08-23 at 15 53 57

### DIFF
--- a/test/jackdaw/admin_test.clj
+++ b/test/jackdaw/admin_test.clj
@@ -143,8 +143,8 @@
     (fn [client]
       (let [{:keys [foo bar]} test-topics]
         (admin/create-topics! client [foo bar])
-        (= {"some-key" "some-value"}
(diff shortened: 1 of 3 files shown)


test/jackdaw/admin_test.clj
```diff
@@ -143,8 +143,8 @@
     (fn [client]
       (let [{:keys [foo bar]} test-topics]
         (admin/create-topics! client [foo bar])
-        (= {"some-key" "some-value"}
-           (admin/get-broker-config client 0))))))
+        (is (= {"some-key" "some-value"}
```
AndreaCrotti: this missing assertion was caught by kaocha

PR #185, opened 2019-08-27
add implementation for all-ok?

AndreaCrotti:


doc/test-machine.md
```diff
@@ -81,10 +81,14 @@ conjunction with `with-open` to ensure that associated resources are shut down
 cleanly when you are finished with a machine.
 
 ```clojure
+(defn all-ok?
```
AndreaCrotti: actually I'm not too sure how this snippet of code is related with the above documentation about Lifecycle, maybe we could move it somewhere or give more context?

AndreaCrotti:
This is not needed anymore if we throw an error, but we should update the documentation in the other PR, since the `all-ok` check is no longer needed.

PR #186, opened 2019-08-27
throw an error if a command failed

AndreaCrotti:
It's not immediately clear when something went wrong, specially in case of timeouts.

Having it failing noisily and printing out the command that failed would help a lot while writing a new test or debugging an existing test.

This is what it would look like if you have an error in one of the commands with this change (with the results also printed out after that as part of the exception).

Screenshot 2019-08-28 at 08 58 27

### DIFF
--- a/doc/test-machine.md
+++ b/doc/test-machine.md
@@ -18,7 +18,7 @@ not write them, and will eventually abandon them)
 The test machine also aims to be agnostic to the system it is testing, other
 than that it should get input/output primarily from Kafka.
 
-### Construction 
+### Construction
 
 The examples below, demonstrate how to create a test machine for executing a
 sequence of "commands". The "local-machine" will execute the commands against a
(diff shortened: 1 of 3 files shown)


99-not-out:
I remember this is how Test Machine used to work, and we changed it deliberately to _not_ fail immediately on the first error due to some incremental fix-one-error-then-the-next-cmd-fails type annoyance. I think the issue was specifically with the `:watch` command - if this times out its sometimes because its in the wrong order, and if the watch is in an intermediate step and you don't let the test run to completion the journal you get back can be less useful in debugging the watcher. So the intention of the current state was to run all the commands and then throw an error at the end if anything failed.

The last bit of that intention (throw an error at the end if anything failed) might have gotten broken at some point though!

So I think everything apart from watcher timeouts _should_ fail fast (a failed write would be fatal say), but the _watcher timeouts_ should not be fatal until the end of all the commands - where we should throw an error if any watcher failed.

AndreaCrotti:
I discussed with @99-not-out but to summarise, it was already previously failing fast on error, and would still not have allowed to run commands after a failed one, so at least now it's failing fast more noisily.

99-not-out:
Approved after chat, we can look at fail fast / not fail fast as optional behaviour in a later change so its explicitly controlled by the caller.

PR #190, opened 2019-09-05
New EDN Serde Without the Newline

creese:
This PR adds a new EDN Serde to a separate location (jackdaw.serdes). In the future, we will be able to require all serdes from here.

This serde fixes a longstanding bug in the EDN serde. It does not print a newline. This is important when writing keys since the presence of a newline causes the record to hash to a different partition when using the default partitioner.

This PR also changes the generated class `jackdaw.serdes.EdnSerde` which now points to the new serde. For applications that define this to be the default key serde, this is a BREAKING CHANGE.

### DIFF
--- a/src/jackdaw/serdes.clj
+++ b/src/jackdaw/serdes.clj
@@ -1,8 +1,72 @@
 (ns jackdaw.serdes
-  "Implements basic SerDes (Serializer/Deserializer)."
+  "Implements string and EDN serdes (serializer/deserializer)."
   (:gen-class)
-  (:import [org.apache.kafka.common.serialization Serdes]))
+  (:require [clojure.edn]
+            [jackdaw.serdes.fn :as jsfn])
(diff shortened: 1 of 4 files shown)


src/jackdaw/serdes.clj
```diff
@@ -1,8 +1,72 @@
...
+  "Returns an EDN deserializer."
+  ([]
+   (edn-deserializer {}))
+  ([opts]
+   (let [opts (into {} opts)]
+     (jsfn/new-deserializer {:deserialize (fn [_ _ data]
+                                            (->> (from-bytes data)
+                                                 (clojure.edn/read-string opts)))}))))
+
+(defn edn-serde
```
cddr: What do you think about making it so that just this bit (i.e. only the function which returns the serde) is in the `jackdaw.serdes` ns and the interop stuff is in some other namespace (suggest jackdaw.serdes.edn2 but not too fussed about the exact name).

PR #191, opened 2019-09-13
Word Count Modernization

creese:
This PR updates the "Word Count" example.

Highlights:
- Removes the ad hoc start/stop functions in favor of James Reeves's Integrant.
- Adds a `dev` alias.
- Adds separate Integrant configs for dev and prod.
- Adds a topic metadata "fake" for use in development.
- Adds a namespace with REPL helpers to create topics, and produce and consume records.
- Copies some parsing logic from the Topology Grapher.

When writing code, you can treat the topic metadata fake just like a map. When used with a "getter", it returns the topic metadata for the name supplied, EDN serdes for the key and value, and a partition count of one.

```
(def topic-metadata (new-fake-topic-metadata))
(:foo topic-metadata)
;; => {:topic-name "foo", 
       :partition-count 1,
       :replication-factor 1,
       :key-serde  #object[org.apache.kafka.common.serialization.Serdes ... ],
       :value-serde #object[org.apache.kafka.common.serialization.Serdes ... ]}
```

When in development, the topics are inferred from the topology. The dev workflow is largely unchanged. Simply jack in and type:
```
(reset)
;; => :reloading (word-count user)
;; => :resumed
(publish (:input topic-metadata) nil "all streams lead to kafka")
(publish (:input topic-metadata) nil "hello kafka streams")
(get-keyvals (:output topic-metadata))
;; => (["all" 1] ["streams" 1] ["lead" 1] ["to" 1] ["kafka" 1] ["hello" 1] ["kafka" 2] ["streams" 2])
```

The `user` namespace is part of the dev alias and is not loaded automatically. To apply the dev alias, add the following to your Emacs config:
```
(setq cider-clojure-cli-global-options "-A:server:client:dev")
```
or type `C-u -A:server:client:dev` when jacking in.


examples/word-count/test/word_count_test.clj
```diff
@@ -1,111 +1,111 @@
...
-                             [(word-watcher "understand")]))
-
-              {:keys [results journal]} (jd.test/run-test machine commands)]
-
-          (is (every? #(= :ok (:status %)) results))
-
-          (is (= 1 (word-count journal "understand")))
-          (is (= 2 (word-count journal "i")))
-          (is (= 3 (word-count journal "to"))))))))
+;; (ns word-count-test
```
creese: I commented out this namespace while development. I plan to put this back.

examples/word-count/dev/user.clj
```diff
@@ -0,0 +1,118 @@
...
+(integrant.repl/set-prep! (constantly config))
+
+
+(deftype FakeTopicMetadata []
+  ILookup
+  (valAt [this key]
+    {:topic-name (name key)
+     :partition-count 1
+     :replication-factor 1
+     :key-serde (js/edn-serde)
```
gphilipp: I think that we should use a string serde for the key to make it easier to join with other topics. See https://github.com/FundingCircle/jackdaw-repl/commit/1ddd902e3ec325c862bdb2ef06f01050255b665e

examples/word-count/dev/user.clj
```diff
@@ -0,0 +1,118 @@
...
+
+(defmethod ig/halt-key! :topics [_ {:keys [streams-config topic-metadata]}]
+  (let [config (word-count/propertize (select-keys streams-config [:bootstrap-servers]))]
+    (re-delete-topics config (re-pattern (str "(" (->> topic-metadata
+                                                       keys
+                                                       (map name)
+                                                       (str/join "|"))
+                                              ")")))))
+
+(defmethod ig/halt-key! :app [_ {:keys [streams-config streams-app]}]
```
gphilipp: I started with that, and then separated out the cleanup or resources because sometimes you just want to stop the app but not delete everything so you can inspect the topics and state stores. https://github.com/FundingCircle/jackdaw-repl/commit/0ba5f6f70ab2a08f8f4d11819c5540d3e093574c

examples/word-count/dev/user.clj
creese: Keep in mind this is just for prototyping. We shouldn't use this or EDN keys in production.

examples/word-count/dev/user.clj
creese: Seems like you'd want a full reset to be able to start clean. Could we halt just the app with a different config structure?

examples/word-count/dev/user.clj
```diff
@@ -0,0 +1,118 @@
...
+     :partition-count 1
+     :replication-factor 1
+     :key-serde (js/edn-serde)
+     :value-serde (js/edn-serde)})
+
+  Associative
+  (assoc [this key val]
+    this))
+
+(defn new-fake-topic-metadata []
```
DaveWM: I think we can just use a single instance of `FakeTopicMetadata`, so we can drop this function and update `topic-metadata` to be:
```clojure
(def topic-metadata (FakeTopicMetadata.))
```
Then we can just use `topic-metadata` where `new-fake-topic-metadata` is currently used.

examples/word-count/dev/user.clj
```diff
@@ -0,0 +1,118 @@
...
+
+(defn destroy-state-stores
+  "Takes an application config and deletes local files associated with
+  internal state."
+  [streams-config]
+  (sh "rm" "-rf" (str "/tmp/kafka-streams/" (:application-id streams-config))))
+
+
+(defmethod ig/init-key :streams-config [_ streams-config]
+  (let [bootstrap-servers (or (System/getenv "BOOTSTRAP_SERVERS")
```
DaveWM: I'd move this logic to where the config map is instantiated, and get rid of this `init-key` method

src/jackdaw/repl.clj
```diff
@@ -0,0 +1,84 @@
...
+  "Takes a topic config, consumes from a Kafka topic, and returns a
+  seq of maps."
+  ([topic-config]
+   (get-records topic-config 200))
+
+  ([topic-config polling-interval-ms]
+   (let [client-config (assoc consumer-config
+                              "group.id"
+                              (str (java.util.UUID/randomUUID)))]
+     (with-open [client (jc/subscribed-consumer client-config [topic-config])]
```
DaveWM: When using this function from the repl, I've always needed to call `seek-to-beginning-eager` on the consumer before using it to consume messages. This may just be my setup though 🤷‍♂️

examples/word-count/dev/user.clj
creese: I was thinking we could add a new arity that takes an argument so the user can set their own default. Would that address your use case?

examples/word-count/dev/user.clj
creese: Could do. Does this still make sense if we add an arity to allow the user to set their own default?

src/jackdaw/repl.clj
creese: This config sets a random UUID for the consumer group. It will always start from the beginning. If we were to stop doing that, I think we would need to do as you suggest. I'm not sure which is preferable.

examples/word-count/src/word_count.clj
```diff
@@ -249,19 +243,22 @@
...
+  (require '[clojure.java.io :as io])
+
   ;; Evaluate the form:
   (let [text-input (slurp (io/resource "metamorphosis.txt"))
         values (str/split text-input #"\n")]
     (doseq [v values]
       (publish (:input topic-metadata) nil v)
-      (info v))
-    (info "The End"))
+      (println v))
```
99-not-out: It might be nice not to print every line, but some other (smaller) progress indicator. I use a fast terminal and it took a while - in an emacs buffer I imagine its a bit painful ...

PR #200, opened 2019-09-23
Transducers are Coming

creese:
This has been a long time coming but I think we’re finally here. This proposal is composable with the existing Jackdaw Streams DSL. Just define your transducers and use `transduce-kstream`:

```
(defn transduce-kstream
  [kstream xf]
  "Takes a kstream and xf and transduces the stream."
  (-> kstream
      (j/transform (fn [] (transformer xf)) ["transducer"])
      (j/flat-map (fn [[_ v]] v))))
```

It turns out that KStream::transform followed by KStream::flatMap is equivalent to transduce with concat. We can use the latter to test our business logic with pure Clojure (no Kafka Streams). This approach was pioneered by Matthias awhile ago. The difference is now we're adding state.

Here is how to test your transducers:
```
(def coll
  [[nil {:debit-account "tech"
         :credit-account "cash"
         :amount 1000}]
   [nil {:debit-account "cash"
         :credit-account "sales"
         :amount 2000}]])

(->> coll
     (transduce (xf-split-entries nil nil) concat)
     (transduce (xf-running-balances (atom {}) swap!) concat))
```

The function `xf-running-balances` takes two arguments, a "store" and a function that "behaves like clojure.core/swap!" and returns a transducer. When developing your tranducers, you can use an atom and swap!.

When using your tranducers from Kafka Streams, no changes are needed. You supply different arguments. The examples show how to provide a state store and a helper function defined in `jackdaw.streams.xform`. However, if this doesn't work for you, you can write your own.

Here is the topology:
```
(require '[jackdaw.streams :as j])
(require '[jackdaw.streams.xform :as jxf])

(defn topology-builder
  [{:keys [entry-requested transaction-pending transaction-added] :as topics} xforms]
  (fn [builder]
    (jxf/add-state-store! builder)
    (-> (j/kstream builder entry-requested)
        (jxf/transduce-kstream (::xf-split-entries xforms))
        (j/through transaction-pending)
        (jxf/transduce-kstream (::xf-running-balances xforms))
        (j/to transaction-added))
    builder))
```

This PR contains examples for Word Count and the Simple Ledger.


kidpollo:
Looking good! I would add examples of unit tests of the actual word count transducer. Also what happens to the simple ledger tests?

src/jackdaw/streams/xform.clj
```diff
@@ -0,0 +1,67 @@
...
+        next (f prev m)]
+    (doall (map (fn [[k v]] (.put store k v)) next))
+    next))
+
+(defn add-state-store!
+  [builder]
+  "Takes a builder and adds a state store."
+  (doto ^StreamsBuilder (j/streams-builder* builder)
+    (.addStateStore (Stores/keyValueStoreBuilder
+                     (Stores/persistentKeyValueStore "transducer")
```
99-not-out: What if you want >1 stateful mapping? won't all the mappers get the same state store? We should perhaps add an airily to pass in the backing store name.

src/jackdaw/streams/xform.clj
```diff
@@ -0,0 +1,67 @@
...
+  org.apache.kafka.streams.kstream.Transformer with overrides for
+  init, transform, and close."
+  [xf]
+  (let [ctx (atom nil)]
+    (reify
+      Transformer
+      (init [_ context]
+        (reset! ctx context))
+      (transform [_ k v]
+        (let [^KeyValueStore store (.getStateStore @ctx "transducer")
```
99-not-out: I would add an airily to pass the backing store name

src/jackdaw/streams/xform.clj
```diff
@@ -0,0 +1,67 @@
...
+              v (first (into [] (xf store) [[k v]]))]
+          (KeyValue/pair k v)))
+      (close [_]))))
+
+(defn transduce-kstream
+  [kstream xf]
+  "Takes a kstream and xf and transduces the stream."
+  (-> kstream
+      (j/transform (fn [] (transformer xf)) ["transducer"])
```
99-not-out: Me again with the backing store name :)

src/jackdaw/streams/xform.clj
```diff
@@ -0,0 +1,67 @@
...
+  [xf]
+  (let [ctx (atom nil)]
+    (reify
+      Transformer
+      (init [_ context]
+        (reset! ctx context))
+      (transform [_ k v]
+        (let [^KeyValueStore store (.getStateStore @ctx "transducer")
+              v (first (into [] (xf store) [[k v]]))]
+          (KeyValue/pair k v)))
```
99-not-out: This transform does not mess with the key - for clarity we should consider providing a `transformer` (which can affect the key) and a `value-transformer` which won't (i.e calls to the underlying `transform` and `transformValues` on the streams DSL).

Similarly for `transduce-kstream` vs `transduce-kstream-values`

This separation is useful when reading the high level code as it telegraphs to the reader whether a repartition may be occurring (a-la `map` vs `mapValues`)

src/jackdaw/streams/xform.clj
```diff
@@ -0,0 +1,67 @@
...
+        (clojure.core/get @store k))
+
+      (put [_ k v]
+        (vswap! store assoc k v)))))
+
+(defn kv-store-swap-fn
+  "Takes an instance of KeyValueStore, a function f, and map m, and
+  updates the store in a manner similar to `clojure.core/swap!`."
+  [^KeyValueStore store f m]
+  (let [ks (keys (f {} m))
```
99-not-out: This (I think) means that the transform must create data when there is none, or else you will get no keys in this step for the subsequent reduction (i.e. it cannot be update-only). Something which should be documented so its explicit - as the result of `(f {} m)` is important. (An update-only transform doesn't really make sense anyway, but worth a note on the doc string IMO)

src/jackdaw/streams/xform.clj
creese: I agree with this. I was planning to add this in a separate PR.

src/jackdaw/streams/xform.clj
creese: Each KStream::transform is always followed by a KStream::flatMap. The final key or keys are obtained from the resulting value of KStream::transform. We don't care about the resulting key. 

Suppose the input to KStream::transform is `[input-key input-value]`. The output might be `[input-key [[k1 v1] [k2 v2]]`. KStream::flatMap discards the key from the previous step and publishes two records with keys k1 and k2.

src/jackdaw/streams/xform.clj
creese: Can you provide a working example?

src/jackdaw/streams/xform.clj
```diff
@@ -0,0 +1,67 @@
...
+              v (first (into [] (xf store) [[k v]]))]
+          (KeyValue/pair k v)))
+      (close [_]))))
+
+(defn transduce-kstream
+  [kstream xf]
+  "Takes a kstream and xf and transduces the stream."
+  (-> kstream
+      (j/transform (fn [] (transformer xf)) ["transducer"])
+      (j/flat-map (fn [[_ v]] v))))
```
cddr: Why is the flat-map necessary here? What if someone wants to keep the keys?

src/jackdaw/streams/xform.clj
cddr: I kind of think this function is the only thing that should really be in this PR (plus a test in xform_test.clj that proves it works). Everything else feels like arbitrary policy decisions that shouldn't be made at the library level.

src/jackdaw/streams/xform.clj
creese: Okay, let me think on that. One `transduce-kstream` could have many stateful transducers. I need to generalize the interface.

src/jackdaw/streams/xform.clj
creese: Flat map corresponds to the reducing function in `transduce`. We always `transduce` with `concat`. This allows you to take a record and transform it stateful into zero or more records. Without this step, you can't implement either of the examples.

src/jackdaw/streams/xform.clj
creese: I disagree. `fake-kv-store` and `kv-store-swap-fn` are what make our transducers reusable. We can test our business logic without the `TopologyTestDriver` and Kafka Streams. `kv-store-swap-fn` allows you to write your transducers as if you were using Clojure atoms. You don't have to use it, but if you want it, it's there. The examples show how to use all this to solve non-trivial problems. I think we need them. I could see adding a few tests though.

src/jackdaw/streams/xform.clj
DaveWM: I think you could achieve the same functionality using the `forward` and `commit` methods on `ctx` (which is a [ProcessorContext](https://kafka.apache.org/10/javadoc/org/apache/kafka/streams/processor/ProcessorContext.html)). There's an example [here](https://github.com/bobby/kafka-streams-clojure/blob/master/src/kafka_streams_clojure/api.clj#L32). Not saying that way's better, just that there is an alternative way.

src/jackdaw/streams/xform.clj
```diff
@@ -0,0 +1,55 @@
...
+        (reset! ctx context))
+      (transform [_ k v]
+        (let [^KeyValueStore store (.getStateStore @ctx "transducer")
+              v (first (into [] (xf store) [[k v]]))]
+          (KeyValue/pair k v)))
+      (close [_]))))
+
+(defn transduce-kstream
+  [kstream xf]
+  "Takes a kstream and xf and transduces the stream."
```
DaveWM: I'd add a bit more information around stateful transducers here. It needs to be clear to the user that this won't work with the stateful transducers from the Clojure core lib, and that if they need a stateful transducer they need to write their own.

src/jackdaw/streams/xform.clj
```diff
@@ -0,0 +1,55 @@
...
+  "Helper functions for working with transducers."
+  (:gen-class)
+  (:require [jackdaw.serdes :as js]
+            [jackdaw.streams :as j])
+  (:import org.apache.kafka.streams.kstream.Transformer
+           org.apache.kafka.streams.KeyValue
+           [org.apache.kafka.streams.state KeyValueStore Stores]
+           org.apache.kafka.streams.StreamsBuilder))
+
+(defn kv-store-swap-fn
```
AndreaCrotti: from the `-fn` suffix I would have thought it would return a function, but it's actually doing the swapping from what I understand.
Maybe something like `default-kv-store-swap` would be more clear?
Do we need the `-swap!` as well if it's a write?

creese: > Maybe something like `default-kv-store-swap` would be more clear?

I like this!


> Do we need the `-swap!` as well if it's a write?

According to the [style guide](https://github.com/bbatsov/clojure-style-guide#changing-state-fns-with-exclamation-mark):
"The names of functions/macros that are not safe in STM transactions should end with an exclamation mark (e.g. reset!)." 


So yeah... I can get behind that.

src/jackdaw/streams/xform.clj
bobby: I was going to make the same point!  I used ProcessorContext in the example @DaveWM linked because it felt closer to a transducible context than anything in the high-level KStreams API.  ProcessorContext also gives a bit more fine-grained control over state stores.

DaveWM:
I spoke to @blak3mill3r (the author of [Noah](https://github.com/blak3mill3r/noah)), yesterday about how he's implemented stateful transducers in Noah. He came up with a broadly similar solution to what we have here, it seems like the use of `volatile!` within the transducer code is a real sticking point. The main difference between our solutions is that in Noah, all the transducers use a single state store rather than each transducer having its own. We weren't sure what the performance implications of that would be, but it's worth bearing in mind in case we run into perf issues in future. 

We also discussed starting a shared library for all the core transducers re-written to support persisting their state, so that they can be used with Jackdaw and Noah. Blake's going to set this up, then I thought we could potentially pull this in in Jackdaw. There are some open questions around this though, such as what we do about other popular transducer libraries like [xforms](https://github.com/cgrand/xforms).

blak3mill3r:
Here is that shared library which reimplements (the transducer arity of) all of the functions in `clojure.core` that return a stateful transducer:

https://github.com/blak3mill3r/coddled-super-centaurs

That function is then bound twice by `noah` to instrument the transducer state and tie it into a `StateStore`:

https://github.com/blak3mill3r/noah/blob/5803dd5/src/noah/transduce.clj#L34-L35
https://github.com/blak3mill3r/noah/blob/5803dd5/src/noah/transduce.clj#L85-L86

Also, @DaveWM ... I checked, and as far as I can tell, there aren't any stateful transducers in [xforms](https://github.com/cgrand/xforms) or [kixi.stats](https://github.com/MastodonC/kixi.stats). They have interesting higher-order transducers and reducing fns, and (I think...) these should work fine composed with these instrumented stateful transducers.

blak3mill3r:
Also I want to clarify regarding: "all the transducers use a single state store rather than each transducer having its own"

Each time you transduce a KStream, if that transduction needs state, you must provide a store. That transducer can of course be a composition of several transducers, any of which can be stateful, and all of the states for these composed transducers will be stored together in a clojure vector as the record in the state store. To transduce multiple KStreams, you would use multiple state stores.

src/jackdaw/streams/xform.clj
```diff
@@ -0,0 +1,55 @@
...
+  init, transform, and close."
+  [xf]
+  (let [ctx (atom nil)]
+    (reify
+      Transformer
+      (init [_ context]
+        (reset! ctx context))
+      (transform [_ k v]
+        (let [^KeyValueStore store (.getStateStore @ctx "transducer")
+              v (first (into [] (xf store) [[k v]]))]
```
vijumathew: if the goal of this line is to get an updated value for `v`, could instead do `(first (sequence xf [[k v]])`. 

i could be missing what `(xf store)` is needed for

src/jackdaw/streams/xform.clj
vijumathew: yeah i think generalizing to handle multiple transducers is good. i think that would handle the implicit coupling with `add-state-store` and `transduce-kstream`. but i imagine there isn't a nice way around the user calling `add-state-store` for each transducer added

src/jackdaw/streams/xform.clj
vijumathew: that makes sense. but might not be clear to someone - if we decide not to replace it with `.forward` and `.commit`, i think this dependency should be documented. or `transformer` becomes a private function so it is clear `transduce-kstream` is the API to use

src/jackdaw/streams/xform.clj
vijumathew: could this just be `(keys m)`?

src/jackdaw/streams/xform.clj
```diff
@@ -0,0 +1,55 @@
...
+  (:require [jackdaw.serdes :as js]
+            [jackdaw.streams :as j])
+  (:import org.apache.kafka.streams.kstream.Transformer
+           org.apache.kafka.streams.KeyValue
+           [org.apache.kafka.streams.state KeyValueStore Stores]
+           org.apache.kafka.streams.StreamsBuilder))
+
+(defn kv-store-swap-fn
+  "Takes an instance of KeyValueStore, a function f, and map m, and
+  updates the store in a manner similar to `clojure.core/swap!`."
```
vijumathew: could we add something that explains the shape of f? maybe like: 

```
f is a function of 2 args that takes the old map (store) and the provided map and combines them together.
```

src/jackdaw/streams/xform.clj
vijumathew: oh wait, is this the reducing function for the state store?

kidpollo:
Plz merge this already 😛 !!

kidpollo:
Bump!

PR #211, opened 2019-11-01
Add an article about jackdaw.admin

AlexVPopov:
This is a first stab at providing a small guide to the `jackdaw.admin` namespace.

# Checklist

- ~[] tests~
- ~[ ] updated [CHANGELOG](../CHANGELOG.md) (the "unreleased" section)~

### DIFF
--- a/doc/admin.md
+++ b/doc/admin.md
@@ -1,0 +1,85 @@
+# Jackdaw AdminClient API
+
+## Rational
+
+The Jackdaw Admin API wraps Kafka's `AdminClient` <sup>[1](#adminapi)</sup> API. It provides functions for obtaining information about a Kafka cluster and managing its topics.
+
+## AdminClient
+
+All functions in the `jackdaw.admin` namespace require a `client`. To produce it, pass the `jackdaw.admin/->AdminClient` function a `config` map. At a minimum, the configuration<sup>[2](#admin-configs)</sup> needs to contain the property `bootstrap.servers` set to the address of a running Kafka broker:
+
+```clojure
+(ns adminclient-example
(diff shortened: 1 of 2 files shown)


doc/admin.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Viewing topics
+
+To simply list the names of the topics in the cluster, use the `list-topics` function:
+
+```clojure
+(ja/list-topics client) ;; => ({:topic-name "jackdaw"})
+```
+
+To view the parition information for a topic, use `describe-topics` function:
```
cddr: s/parition/partition/

doc/admin.md
```diff
@@ -0,0 +1,85 @@
...
+;;                                                                        :name "cleanup.policy",
+;;                                                                        :read-only? false,
+;;                                                                        :sensitive? false,
+;;                                                                        :value "compact"},
+;;                                                                                     ...}}
+```
+
+### Changing topics' configuration
+
+To change the configuration of a topic, pass the function `alter-topic-config!` a client and an array of topic descriptor maps. Each map needs to contain two keys - the `:topic-name`, as well as a `:topic-config`, holding the configuration<sup>[3](#topic-configs)</sup> of the topic as stringified key-value pairs.
```
cddr: Maybe consider switching the "array" in this line for "vector"?

AlexVPopov: Absolutely! Coming from Ruby I always tend to say `array` and `hash`, instead of `vector` and `map`, well spotted.

AlexVPopov:
@cddr PR updated.

PR #219, opened 2019-12-20
Add documentation for cutting releases of jackdaw

cddr:
Now that the release script is working properly, the instructions for cutting a release are quite simple. This PR adds them to the README.


AndreaCrotti:
Looks good, we could mention also that it's really the same as just doing a `git tag $version` and `git push --tags` locally but apart from that it's fine

PR #224, opened 2020-01-30
Open up the coercion stack

cddr:
A few people have noticed that the **JSON** representation of an avro object is a bit different to the edn representation of an avro object. This behaviour was originally copied from the [abracad.avro project](https://github.com/damballa/abracad) but it makes it a little bit inconvenient to work with data obtained via more "avro-oriented" tools like the [kafka-avro-console-consumer](https://github.com/confluentinc/schema-registry/blob/master/bin/kafka-avro-console-consumer) or the [avro-random-generator](https://github.com/confluentinc/avro-random-generator).

The pair of functions in this PR make it a bit easier to convert between these two representations.

### as-json
```
(as-json {:avro-schema (clojure.data.json/write-str ["null" "string"])}
          "yolo")
=> "{\"string\":\"yolo\"}"
```

### as-edn
```
(as-edn {:avro-schema (clojure.data.json/write-str ["null" "string"])}
        "{\"string\":\"yolo\"}")
=> "yolo"
```

## Checklist

- [x] factor out the common code that builds the coercion-stack
- [x] tests

### DIFF
--- a/test/jackdaw/serdes/avro_test.clj
+++ b/test/jackdaw/serdes/avro_test.clj
@@ -557,18 +557,21 @@
                                   :type-registry (merge avro/+base-schema-type-registry+
                                                         avro/+UUID-type-registry+)}
                                  valid-edn)
-                   (json/read-str))]
-      (is (= "a_1" (get json "enum_field")))
-      (is (= nil (get json "nil_field")))
-      (is (= {"map" {"banana" {"color" "yellow"}, "ripe b4nana$" {"color" "yellow-green"}}}
-             (get json "map_field")))
-      (is (= 1 (get json "default_field")))
-      (is (= "00000000-0000-0000-0000-000000000000" (get json "uuid_field")))
-      (is (= {"array" [{"color" "yellow"}]}
-             (get json "array_field")))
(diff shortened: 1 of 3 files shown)


test/jackdaw/serdes/avro_test.clj
```diff
@@ -515,6 +515,61 @@
...
+                     :uuid-field #uuid "00000000-0000-0000-0000-000000000000",
+                     :array-field [{:color "yellow"}],
+                     :string-field "hello",
+                     :optional-field nil,
+                     :long-field 3}
+          json (-> (avro/as-json {:avro-schema complex-schema-str
+                                  :type-registry (merge avro/+base-schema-type-registry+
+                                                        avro/+UUID-type-registry+)}
+                                 valid-edn)
+                   (json/read-str))]
```
AndreaCrotti: if you add a `keywordize-keys` here (or probably pass an option to json/read-string) you can avoid all the `get` below and just use the usual keyword syntax.

AndreaCrotti:
> No tests yet.

You can remove that from the description of the PR since you added them

PR #221, opened 2020-01-17
add link to the topology grapher

AndreaCrotti:

PR #225, opened 2020-02-03
delete code that was copied from the topology-grapher

AndreaCrotti:
Now the topology-grapher is also OSS, if we want to help allow people
to graph their topologies easily they can simply do that adding another
dependency to their dev profiles.

# Checklist

- [ ] tests


AndreaCrotti:
Might be good to merge this first https://github.com/FundingCircle/jackdaw/pull/221
so there is a reference about how do this when it's deleted from here.

dev/jackdaw/repl.clj
```diff
@@ -115,21 +114,6 @@
   given with EDN serdes and a partition count of one."
   (FakeTopicMetadata.))
 
-(defn topology->topic-metadata
```
creese: ☝️ This function infers the Kafka topics needed by a topology. We can't simply delete it as this will break examples.

AndreaCrotti: Ah sorry I thought this ns was only used in the repl.
Can we just add the grapher as dependency to the examples?

creese: That is one option.

The namespace is referenced here:
https://github.com/FundingCircle/jackdaw/blob/master/examples/word-count/dev/user.clj#L32

AndreaCrotti: Ok thanks I'll fix the examples then

PR #231, opened 2020-02-13
Tighten up the test-machine specs

cddr:
In #230 we added helper functions for creating test-commands which helps in cases where the commands are authored using these new functions. @AndreaCrotti commented that it would be nice if we checked that the commands passed to `run-test` did indeed match the specs. This PR includes specs for some of the legacy commands left out of that PR so that we can now properly assert the test-command sequences passed to run-test.

To enable spec-checking behaviour during development evaluate `(s/check-asserts true)` at the REPL. When this is enabled and you attempt to `run-test` with an invalid command sequence, you'll get an exception like so

Screen Shot 2020-02-13 at 4 27 11 PM

# Checklist

- [x] tests


test/jackdaw/test/commands_test.clj
```diff
@@ -16,3 +17,45 @@
     (is (thrown-with-msg? clojure.lang.ExceptionInfo
                           #"Unknown command: :not-found"
                           (cmd/command-handler {} [:not-found])))))
+
+(defn valid-command?
+  [cmd]
+  (not (= :clojure.spec.alpha/invalid
+          (s/conform ::cmd/test-event cmd))))
```
minimal: or just `(s/valid? ::cmd/test-event cmd)`

PR #239, opened 2020-04-02
standalone deser

noisesmith:
Updates some stale comments.

Changes the spec for avro serdes so that the schema document is optional. This allows a reader to continue using a topic as the producer evolves their schema, without updating the schemas documents in lockstep.

# Checklist

- [x] tests

### DIFF
--- a/src/jackdaw/serdes/avro.clj
+++ b/src/jackdaw/serdes/avro.clj
@@ -10,26 +10,26 @@
   The intentional API of this NS has three main features -
   `SchemaCoercion`, the intentional type registry (of which
   `#'+base-schema-type-registry+` is an example) and
-  `#'avro-serde`.
+  `#'serde`.
 
-  `avro-serde` is the primary entry point to this namespace for
-  users. It's a function of a schema-registry configuration, a schema
+  `serde` is the primary entry point to this namespace for users.
+  It's a function of a schema-registry configuration, a schema
   type registry, and a serde configuration to be instantiated.
 
-  The intent is that an end user will `partial` the `avro-serde`
+  The intent is that an end user will `partial` the `serde`
   function with their schema registry details and desired type
   registry, and use the `partial`'d function as en entry in a registry
   as used by `jackdaw.serdes/serde`.
 
-  This allows `serdes` and `avro-serde` to be agnostic to application
-  or environment specific configuration details.
+  This allows `serdes` and `serde` to be agnostic to application or
+  environment specific configuration details.
 
   But what's this type registry?
 
   Apache Avro \"logical types\" - a tool for annotating fields in an
   avro record as having some complex interpretation beyond their
   serialized format. The `type-registry` for the purposes of the
-  `avro-serde` function a mapping of addresses to functions which will
+  `serde` function a mapping of addresses to functions which will
   when invoked build and return a `SchemaCoercion` instance.
 
   When a Serde is instantiated, a stack of `SchemaCoercion` coersion
@@ -50,7 +50,7 @@
   appropriate handler.
 
   A user who wanted to opt into these handlers could simply call
-  `avro-serde` with
+  `serde` with
   `(merge +base-schema-type-registry+ +UUID-type-registry+)`
 
   Users are HIGHLY encouraged to use the `+base-schema-type-registry+`
--- a/src/jackdaw/serdes/avro.clj
+++ b/src/jackdaw/serdes/avro.clj
@@ -505,7 +505,7 @@
   {"schema.registry.url" registry-url})
 
 (defn- serializer [schema->coercion serde-config]
-  (let [{:keys [registry-client registry-url avro-schema key?]} serde-config
+  (let [{:keys [registry-client registry-url avro-schema read-only? key?]} serde-config
         base-serializer (KafkaAvroSerializer. registry-client)
         ;; This is invariant across subject schema changes, shockingly.
         coercion-type (schema->coercion avro-schema)
@@ -514,6 +514,9 @@
                  :configure (fn [_ base-config key?]
                               (.configure base-serializer base-config key?))
                  :serialize (fn [_ topic data]
+                              (when read-only?
+                                (throw (ex-info "Cannot serialize from a read-only serde"
+                                                {:serde-config serde-config})))
                               (try
                                 (.serialize base-serializer topic (clj->avro coercion-type data []))
                                 (catch clojure.lang.ExceptionInfo e
@@ -671,7 +674,8 @@
    {:keys [avro/schema
            avro/coercion-cache
            key?
-           deserializer-properties]
+           deserializer-properties
+           read-only?]
     :as   topic-config}]
 
   (when-not url
@@ -691,6 +695,7 @@
                                      (registry/client url 128))
                 ;; Provide the old behavior by default, or fall through to the
                 ;; new behavior of getting the right schema when possible.
+                :read-only?      read-only?
                 :avro-schema     (parse-schema-str schema)}
 
         coercion-stack {:type-registry type-registry
--- a/src/jackdaw/serdes/avro/confluent.clj
+++ b/src/jackdaw/serdes/avro/confluent.clj
@@ -8,7 +8,8 @@
   serde operation."
   [schema-registry-url schema key? & [{:keys [type-registry
                                               schema-registry-client
-                                              coercion-cache]}]]
+                                              coercion-cache
+                                              read-only?]}]]
   (let [reg (if (nil? type-registry)
               jsa/+base-schema-type-registry+
               type-registry)]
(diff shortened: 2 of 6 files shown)


src/jackdaw/specs.clj
```diff
@@ -32,9 +32,11 @@
...
 
 (s/def :jackdaw.serde/confluent-avro-serde
-  (s/keys :req-un [::serde-keyword
-                   ::key?
-                   (or ::schema ::schema-filename)]))
+  (s/keys
+    :opt-un [::schema
+             ::schema-filename]
+    :req-un [::serde-keyword
+             ::key?]))
```
MicahElliott, line 39: Maybe worth have the `req-un` before that `opt-un`. I feel I see that most commonly.

noisesmith:
I'm still working on the test coverage and CHANGELOG update

creese:
We may want to create a separate top-level function for when a deserializer-only serde is wanted. That feels a bit safer to me than relaxing the current spec.

src/jackdaw/specs.clj
noisesmith: 👍

noisesmith:
> We may want to create a separate top-level function for when a deserializer-only serde is wanted. That feels a bit safer to me than relaxing the current spec.

Thanks Charles, I think the current version of the PR does what you want.

src/jackdaw/serdes/avro/confluent.clj
```diff
@@ -18,3 +18,16 @@
                {:avro/schema schema
                 :key? key?
                 :avro/coercion-cache coercion-cache})))
+
+(defn reader-serde
+  "Creates a serde for Avro data with an optional schema.
```
creese, line 23: This shouldn't take a schema.

src/jackdaw/specs.clj
```diff
@@ -36,6 +36,13 @@
                    ::key?
                    (or ::schema ::schema-filename)]))
 
+;; for a serde that is only used for reading, using the schema from the registry
+(s/def :jackdaw.serde/confluent-avro-reader-serde
+  (s/keys
+    :req-un [::serde-keyword
+             ::key?]
+    :opt-un [::schema
+             ::schema-filename]))
```
creese, line 45: `::key?` is only needed for the Schema Registry. We can probably reuse `:jackdaw.serde/serde` defined above.

src/jackdaw/serdes/avro/confluent.clj
```diff
@@ -18,3 +18,16 @@
...
+
+  See `serde`.
+
+  serde itself can work (for reading only) without providing a schema.
+  This function is needed because the resolver uses the function key to derive
+  the spec to enforce for its arguments, and we don't want `serde` to use
+  this looser spec (see :jackdaw.serde/confluent-avro-reader-serde and
+  :jackdaw.serde/confluent-avro-serde)."
+  [schema-registry-url schema key? & [opts-map]]
+  (serde schema-registry-url schema key? opts-map))
```
creese, line 33: This should call `jsa/serde` without a schema.

src/jackdaw/serdes/resolver.clj
```diff
@@ -54,12 +54,16 @@
...
+            spec-mapping {:jackdaw.serdes.avro.confluent/serde :jackdaw.serde/confluent-avro-serde
+                          :jackdaw.serdes.avro.confluent/reader-serde  :jackdaw.serde/confluent-avro-reader-serde}
+            spec-key (spec-mapping serde-keyword)]
+        (case spec-key
+          (:jackdaw.serde/confluent-avro-serde
+           :jackdaw.serde/confluent-avro-reader-serde)
+          (if-not (s/valid? spec-key serde-config)
             (throw (ex-info "Invalid serde config."
-                            (s/explain-data :jackdaw.serde/confluent-avro-serde serde-config)))
+                            (s/explain-data spec-key serde-config)))
```
creese, line 66: The serde-resolver requires a schema. Can we back out these changes?

noisesmith: it's only the serialization side that uses it, and the current kafka-productionised code looks up everything using the resolver

An alternate approach would be to allow the resolver (or something with a similar API) to return a wrapper with only the `deserializer` field.

The current approach is compatible with the existing code paths because the expected deserializer field is present and the resolver adds the expected conversions on read, I'm open to other options but would like to limit the scope of the API change. What about creating a wrapper where the `deserializer` works via the schema registry and the `serializer` simply throws an informative error?

cddr:
I thought that this PR https://github.com/FundingCircle/jackdaw/pull/157 meant that `schema` was no longer required but looking back, I don't see a test that proves it to be fair.

src/jackdaw/serdes/resolver.clj
```diff
@@ -54,12 +54,16 @@
       (let [schema (cond
                      schema-filename (load-schema serde-config)
                      schema          schema)
```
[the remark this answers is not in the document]
noisesmith: `cond` just works that way, but I added an explicit nil for clarity

```
(ins)user=> (cond true 1)
1
(ins)user=> (cond false 1)
nil
```

noisesmith:
> I thought that this PR #157 meant that `schema` was no longer required but looking back, I don't see a test that proves it to be fair.

I think this is true - but the spec used by the resolver would still fail, and I think I want that entry point.

test/jackdaw/serdes/resolver_test.clj
```diff
@@ -98,6 +98,29 @@
...
+    (testing "avro reader serdes with optional schema"
+      (let [resolver-fn (resolver/serde-resolver :schema-registry-url ""
+                                                 :schema-registry-client (reg/mock-client))
+            writer-avro-config {:serde-keyword :jackdaw.serdes.avro.confluent/serde
+                                :schema-filename "resources/example_schema.avsc"
+                                :key? false}
+            ;; by providing :serde/type of :read-only we can allow a serde without local schema
+            reader-avro-config {:serde-keyword :jackdaw.serdes.avro.confluent/serde
+                                :key? false
+                                :serde-type :read-only}
```
creese, line 110: I would rather have this work like `:key?`.

Example:
```
{:serde-keyword :jackdaw.serdes.avro.confluent/serde
 :key? false
 :read-only? true}
```

Or if `:key?` is not required:
```
{:serde-keyword :jackdaw.serdes.avro.confluent/serde
 :read-only? true}
```

creese:
What happened to `jackdaw.serdes.avro.confluent/reader-schema`?

src/jackdaw/serdes/resolver.clj
```diff
@@ -39,6 +39,8 @@
   Options:
   schema-registry-url - The URL for the schema registry
   type-registry - A mapping per jackdaw.serdes.avro/+base-schema-type-registry+>
+  read-only - Specifies that you will not be using the resulting serializer,
+              and does not require a schema or schema-filename
```
99-not-out, line 43: What will happen if you try to write with a read-only avro serdes?
Would be nice to get a clean error message, as debugging avro error is a black art at the best of times!

noisesmith: I've added an informative message, and a test of that message.

src/jackdaw/specs.clj
```diff
@@ -31,11 +31,21 @@
...
   (s/keys :req-un [::serde-keyword
                    ::key?
                    (or ::schema ::schema-filename)]))
 
+(defmethod avro-serde-type :read-only [_]
+  (s/keys :req-un [::serde-keyword
+                   ::key?]))
+
+(s/def :jackdaw.serde/confluent-avro-serde
+  (s/multi-spec avro-serde-type :read-only?))
```
creese, line 48: Can we extend the existing key group?
```
(s/def :jackdaw.serde/confluent-avro-serde
  (s/keys :req-un [::serde-keyword
                             ::key?
                             (or ::read-only? ::schema ::schema-filename)])
```

src/jackdaw/serdes/avro.clj
```diff
@@ -671,7 +674,8 @@
    {:keys [avro/schema
            avro/coercion-cache
            key?
-           deserializer-properties]
+           deserializer-properties
```
creese, line 677: What is this?

src/jackdaw/specs.clj
```diff
@@ -31,11 +32,20 @@
...
+(s/def :jackdaw.serde/confluent-avro-serde
+  (s/and
+   (s/keys :req-un [::serde-keyword
+                    ::key?
+                    (or ::schema
+                        ::schema-filename
+                        ::read-only?)])
+   #(exactly-one-true? (string? (:schema %))
+                       (string? (:schema-filename %))
+                       (true? (:read-only? %)))))
```
creese, line 48: I thought we agreed just to check for the key's presence or absence.

src/jackdaw/serdes/avro.clj
noisesmith: it pre-exists this pr, introduced in commt #8551c18c836b7b52e0a0b16b9b041d30b3eb4a0c

src/jackdaw/specs.clj
noisesmith: I misunderstood - easy enough to fix

PR #245, opened 2020-04-27
Add flat-transform wrappers for Kafka Streams

99-not-out:
Spotted [flatTransform](https://kafka.apache.org/23/javadoc/org/apache/kafka/streams/kstream/KStream.html#flatTransform-org.apache.kafka.streams.kstream.TransformerSupplier-java.lang.String...- ) for a KStream and it looks worth adding. Also added a few of the common bits of wrapper / helper / sugar code which pop up from uses of transformers.

* Add support for flat-transform and flat-transform-values
* Add some helpers for adding state stores
* Add some sugar for common transform use cases
* Use some of the above to provide simple de-duping stream support in extras

With `flatTransformValues` note that if you return an empty list or nil, it halts processing. With `transformValues`, if you return nil processing is *not* halted and the next processor in the stream sees [k nil]. This difference can be useful if you _meant_ to halt processing, as it avoids a subsequent filter.

@creese We really need to get your transducers changes merged too!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -122,8 +122,6 @@ jobs:
           lein kaocha --plugin kaocha.plugin/junit-xml --junit-xml-file test-results/kaocha/results.xml
       - store_artifacts:
           path: ./logs
-      - store_artifacts:
-          path: target/coverage
       - store_test_results:
           path: test-results
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -119,11 +119,16 @@ jobs:
       - setup_remote_docker:
           docker_layer_caching: true
       - run:
-          lein kaocha --plugin kaocha.plugin/junit-xml --junit-xml-file test-results/kaocha/results.xml
+          lein kaocha --plugin kaocha.plugin/junit-xml --junit-xml-file test-results/kaocha/results.xml --plugin cloverage --codecov
       - store_artifacts:
           path: ./logs
+      - store_artifacts:
+          path: target/coverage
       - store_test_results:
           path: test-results
+      - run:
+          name: Report Test Coverage
+          command: bash <(curl -s https://codecov.io/bash)
 
   deploy:
     <<: *deploy_config
--- a/src/jackdaw/streams.clj
+++ b/src/jackdaw/streams.clj
@@ -46,6 +46,12 @@
   [streams-builder]
   (p/source-topics streams-builder))
 
+(defn with-state-store
+  "Adds a persistent state store to the topology with the configured name
+  and serdes"
+  [streams-builder store-config]
+  (p/with-state-store streams-builder store-config))
+
 (defn streams-builder*
   "Returns the underlying KStreamBuilder."
   [streams-builder]
@@ -200,6 +206,15 @@
   ([kstream transformer-supplier-fn state-store-names]
    (p/transform kstream transformer-supplier-fn state-store-names)))
 
+(defn flat-transform
+  "Creates a KStream that consists of the results of applying the transformer
+  to each value in the input stream. Result of the transform should be iterable,
+  and the resulting stream is as per flatMap"
+  ([kstream transformer-supplier-fn]
+   (p/flat-transform kstream transformer-supplier-fn))
+  ([kstream transformer-supplier-fn state-store-names]
+   (p/flat-transform kstream transformer-supplier-fn state-store-names)))
+
 (defn transform-values
   "Creates a KStream that consists of the results of applying the transformer
   to each value in the input stream."
@@ -207,6 +222,15 @@
    (p/transform-values kstream value-transformer-supplier-fn))
   ([kstream value-transformer-supplier-fn state-store-names]
    (p/transform-values kstream value-transformer-supplier-fn state-store-names)))
+
+(defn flat-transform-values
+  "Creates a KStream that consists of the results of applying the transformer
+  to each value in the input stream. Result of the transform should be iterable,
+  and the resulting stream is as per flatMap"
+  ([kstream value-transformer-supplier-fn]
+   (p/flat-transform-values kstream value-transformer-supplier-fn))
+  ([kstream value-transformer-supplier-fn state-store-names]
+   (p/flat-transform-values kstream value-transformer-supplier-fn state-store-names)))
 
 (defn join-global
   [kstream global-ktable kv-mapper joiner]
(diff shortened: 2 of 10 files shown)


99-not-out:
Coverage doesn't seem to like `jackdaw/streams/protocols.clj` anymore 😞

.circleci/config.yml
```diff
@@ -118,16 +118,11 @@ jobs:
       - setup_remote_docker:
           docker_layer_caching: true
       - run:
-          lein kaocha --plugin kaocha.plugin/junit-xml --junit-xml-file test-results/kaocha/results.xml --plugin cloverage --codecov
+          lein kaocha --plugin kaocha.plugin/junit-xml --junit-xml-file test-results/kaocha/results.xml
```
99-not-out, line 121: Removed coverage check for now as it fails on the `interop` namespace.

src/jackdaw/streams.clj
```diff
@@ -200,6 +215,15 @@
   ([kstream value-transformer-supplier-fn state-store-names]
    (p/transform-values kstream value-transformer-supplier-fn state-store-names)))
 
+(defn flat-transform-values
```
kidpollo, line 218: why not `flat-transform-keys` ?

src/jackdaw/streams/interop.clj
```diff
@@ -323,7 +333,20 @@
     (clj-kstream
      (.transform ^KStream kstream
                  ^TransformerSupplier (transformer-supplier transformer-supplier-fn)
-                 ^"[Ljava.lang.String;" (into-array String state-store-names))))
+                 ^"[Ljava.lang.String;" (into-array String
```
kidpollo, line 336: is that how a string array is annotated? 🤯

test/jackdaw/streams_test.clj
```diff
@@ -443,6 +444,101 @@
...
+
+  (testing "with-state-store and transform-with-ctx"
+    (let [topic-a (mock/topic "topic-a")
+          topic-b (mock/topic "topic-b")
+          driver (mock/build-driver (fn [builder]
+                                      (-> builder
+                                          (k/with-state-store {:store-name "test-store"})
+                                          (k/kstream topic-a)
+                                          (k/transform
+                                           (lambdas/transformer-with-ctx
```
kidpollo, line 490: Wondering if a macro could make this feel more 'clojurey' like 
```clojure
(transformer-with-context [test-store k v]
(update test-store k some-fn)
)
```

src/jackdaw/streams/interop.clj
99-not-out: Yeah, although arrays are objects in Java they are not an instance of a class, so have a slightly funky type signature :-)

test/jackdaw/streams_test.clj
99-not-out: Yeah there is possible a macro here as the state store names have to be repeated. I try to avoid reaching for macro's immediately however until the basic functions are there.

We could perhaps start a macros namespace, as there are probably several syntax things you could add over the basic functions for use in streams.

src/jackdaw/streams/configured.clj
```diff
@@ -52,6 +52,10 @@
     [_]
     (source-topics streams-builder))
 
+  (with-state-store
```
99-not-out, line 55: I am not 100% on the naming here - maybe `add-state-store`. However this name make the code using it read nicer IMHO. `(-> builder (k/with-state-store ...) (k/stream ...))` etc.

src/jackdaw/streams/interop.clj
```diff
@@ -314,6 +324,26 @@
...
+    (clj-kstream
+     (.leftJoin kstream
+                ^GlobalKTable (global-ktable* global-ktable)
+                ^KeyValueMapper (select-key-value-mapper key-value-mapper-fn)
+                ^ValueJoiner (value-joiner joiner-fn))))
+
+  (kstream* [_]
+    kstream)
+
+  ITransformingKStream
```
99-not-out, line 346: Added a new protocol for the transform fns on a stream, which is always used wherever IKStream is used. This had to be done to allow Coverage to run over this IKStream protocol.

Its not ideal, and perhaps there is a  newer / different code coverage tool to fix this.

The change is however invisible to users of Jackdaw.

kidpollo:
This PR and the #305 seem to have related/similar changes. @99-not-out Is there any insight on which will win hahaha or if both would be merged. I kinda want some of these stuff. Would love this merged and a new release cut sometime soon :P

kidpollo:
@99-not-out would it make sense to add a simple interop to set message headers?

Right now one would have to do something like:

``` clojure 
...

(defn set-headers [context headers]
  (let [record-headers (.headers context)]
    (doseq [[k v] headers]
      (.add record-headers
            k
            (.getBytes v))))
  context)
...
(-> stream
          (j/transform-values
           (jl/value-transformer-with-ctx
            (fn [ctx v]
              (set-headers ctx {"foo" "bar"})
              v)))
```

99-not-out:
Going to close this PR in favour of https://github.com/FundingCircle/jackdaw/pull/305
(which takes the core parts of this PR and will hopefully me merged soon!)

99-not-out:
> This PR and the #305 seem to have related/similar changes. @99-not-out Is there any insight on which will win hahaha or if both would be merged. I kinda want some of these stuff. Would love this merged and a new release cut sometime soon :P

#305 will win :-)

99-not-out:
> @99-not-out would it make sense to add a simple interop to set message headers?
> 
> Right now one would have to do something like:
> 
> ```clojure
> ...
> 
> (defn set-headers [context headers]
>   (let [record-headers (.headers context)]
>     (doseq [[k v] headers]
>       (.add record-headers
>             k
>             (.getBytes v))))
>   context)
> ...
> (-> stream
>           (j/transform-values
>            (jl/value-transformer-with-ctx
>             (fn [ctx v]
>               (set-headers ctx {"foo" "bar"})
>               v)))
> ```

I did have a bit of a think about headers a while a go - there is even a PR somewhere I think to add support in test machine.

Lets start a new PR specifically for headers @kidpollo  -  there are some nuances in there (like supporting encodings, so the interop isn't all `byte[]` based ...). Will tag you wen we start one (or feel free to start one yourself!)

gphilipp:
Superseded by https://github.com/FundingCircle/jackdaw/pull/305/files.

Issue #258, opened 2020-06-15
Type union coercion fails when record types come before others


dacreify:
# Summary

The `match-clj?` implementation on `RecordType` doesn't check if `clj-map` is actually a map, which it may not be if part of a type union.

# Repro

1. Go into `jackdaw.serdes.avro-test/schema-type-test` to the `(testing "union"` form
1. In the `avro-schema` binding move `record-1-schema` so that it's the first element of the vector
1. Run the test and observe that it fails

PR #259, opened 2020-06-15
Type union coercion fails when record types come before others [#258]

dacreify:
# Checklist

- [x] tests

### DIFF
--- a/src/jackdaw/serdes/avro.clj
+++ b/src/jackdaw/serdes/avro.clj
@@ -389,15 +389,17 @@
 (defrecord RecordType [^Schema schema field->schema+coercion]
   SchemaCoercion
   (match-clj? [_ clj-map]
-    (let [[_ unknown-fields _] (clojure.data/diff (set (keys field->schema+coercion))
-                                                  (set (keys clj-map)))]
-      (and (every? (fn [[field-key [^Schema$Field field field-coercion]]]
-                     (let [field-value (get clj-map field-key ::missing)]
-                       (if (= field-value ::missing)
-                         (.defaultValue field)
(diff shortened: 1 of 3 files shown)


src/jackdaw/serdes/avro.clj
```diff
@@ -389,15 +389,17 @@
...
-    (let [[_ unknown-fields _] (clojure.data/diff (set (keys field->schema+coercion))
-                                                  (set (keys clj-map)))]
-      (and (every? (fn [[field-key [^Schema$Field field field-coercion]]]
-                     (let [field-value (get clj-map field-key ::missing)]
-                       (if (= field-value ::missing)
-                         (.defaultValue field)
-                         (match-clj? field-coercion field-value))))
-                   field->schema+coercion)
-           (empty? unknown-fields))))
+    (if (map? clj-map)
```
cddr, line 392: Since the `else` clause returns false, I think this could maybe just be `(when (map? clj-map) ...)`.

cddr (commented):
LGTM. Only a minor stylistic suggestion.

src/jackdaw/serdes/avro.clj
dacreify: @cddr That's what I had initially, but then I noticed explicit `(boolean ..` forms in other `match-clj?` implementations, and so thought maybe returning falsey `nil` was against the contract. If not glad to change!

dacreify: Made this change and also updated `CHANGELOG`. Tried to re-request review but the button isn't doing anything. 🤷🏼‍♂️

PR #260, opened 2020-06-21
Add kafka message header support for test machine

99-not-out:
Adds basic support to test machine for reading and writing message headers.
Unfortunately, the Kafka Rest API does not appear to support message headers so its not available there - however the broker and mock transports do support headers.

This is deliberately a light touch Test Machine only change to allow testing and experimentation with the support of headers more widely in Jackdaw, or any topologies tested via Test Machine. No attempt at serdes support is added here, headers must be provided in the form the Java API expects which is a map of:

`String k` -> `byte[] value`

The support does at least let you provide them in a clojure Map. Writes are supported via an additional option for the `:write!` command, e.g.:

`[:write! "my-topic" {:id 1 :message "yolo"} {:headers {"VERSION" (.getBytes "1.0.0")}]`

Any headers read into the journal for a test are captured in the top level structure for each message in the journal, e.g. for the above write you would see:

```
{:topic "my-topic",
 :key 1,
 :value  {:id 1, :message "yolo"},
 :partition 0,
 :offset 0,
 :headers {"VERSION" ...byte array...}}  <<-- headers added here
```

It is left to the user to encode and decode the `byte[]`'s used for headers (as per the Kafka API).

See `test-write-then-read-with-headers ` in `test/jackdaw/test_test.clj` for an end to end example.

### DIFF
--- a/src/jackdaw/test/transports/mock.clj
+++ b/src/jackdaw/test/transports/mock.clj
@@ -29,9 +29,15 @@
 ;; For this reason, we try to make things a bit more meaningful by using
 ;; terms like "input-record" and "output-record".
 
+(defn set-headers [consumer-record headers]
+  (let [record-headers (.headers consumer-record)]
+    (doseq [[k v] headers]
+      (.add record-headers k v)))
+  consumer-record)
+
 (defn with-input-record
   "Creates a kafka ConsumerRecord to be fed directly into a topology source
-   node by the TopologyTestDriver"
+  node by the TopologyTestDriver"
   [_topic-config]
   (fn [m]
     (let [record (ConsumerRecord. (get-in m [:topic :topic-name])
@@ -48,7 +54,8 @@
                                     0)
                                   (:key m)
                                   (:value m))]
-     (assoc m :input-record record))))
+      (set-headers record (:headers m))
+      (assoc m :input-record record))))
 
 (defn with-output-record
   [_topic-config]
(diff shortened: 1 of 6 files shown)


src/jackdaw/test/transports/mock.clj
```diff
@@ -56,7 +56,12 @@
     {:topic (.topic r)
      :key (.key r)
      :value (.value r)
-     :partition (or (.partition r) -1)}))
+     :partition (or (.partition r) -1)
+     :matt-was-here true
```
cddr, line 60: Presume this was left-over from testing? :-)

LGTM otherwise.

99-not-out: Doh! yes :-)
Hope you are doing well Andy!

marcoccchan:
Can we document the new `:headers` option in https://github.com/FundingCircle/jackdaw/blob/8cbc24d89fef2219a9fc0d7dc94a04fceac4cbf0/src/jackdaw/test/commands.clj#L80-L86?

src/jackdaw/test/transports/rest_proxy.clj
```diff
@@ -210,6 +210,7 @@
   [config topic-metadata deserializers]
   (let [continue?   (atom true)
         xform       (comp
+                     #(assoc % :headers {}) ; cannot read headers over the Rest API
```
AndreaCrotti, line 213: do we need to pass it empty if it's not supported yet or can we just not set it?

99-not-out: It can be just not set - I will add some docs though!

PR #266, opened 2020-07-20
Restore the fixtures module to the public interface

brett-lempereur:
This module was part of the public interface of the library, and we depend on it in our integration test suites.


cddr:
As far as I can tell, it was this PR by @gphilipp that changed things

https://github.com/FundingCircle/jackdaw/pull/235

Seems the main purpose of that PR was to remove the dependency on the kafka server which seems reasonable. Was there a reason for also removing the fixtures from the public interface? Is there a workaround?

99-not-out (dismissed):

99-not-out:
There was a change made to avoid the need of pulling in the admin interface for Kafka (as it hoiks in Scala etc, and makes the  base  dependencies fairly complex as a result). This involved not putting the `test/` tree in the published JARs.
The fixtures NS is very useful though - its fine to move it under src, however no-one will be able to use it without adding the required deps themselves under their test profile - so we should add a note to this effect.

brett-lempereur:
> There was a change made to avoid the need of pulling in the admin interface for Kafka (as it hoiks in Scala etc, and makes the base dependencies fairly complex as a result). This involved not putting the `test/` tree in the published JARs.
> The fixtures NS is very useful though - its fine to move it under src, however no-one will be able to use it without adding the required deps themselves under their test profile - so we should add a note to this effect.

Thanks, would the best place for that be in the `README.md`, or the documentation for the `jackdaw.test.fixture` module?

gphilipp:
> ... This involved not putting the `test/` tree in the published JARs.

Well, the `test/` tree is never meant to be included in the published artefact 💁‍♂️. Re-reading the original PR, I realise it's a 
mistake that the fixture ns ended up being moved into the `test` directory, as the change didn't require it. We discussed it internally at some point, decided it was not necessary yet the move went in 🤦 

> Thanks, would the best place for that be in the README.md, or the documentation for the jackdaw.test.fixture module?

I think the best place would be the README.

cddr:
> Re-reading the original PR, I realise it's a
mistake that the fixture ns ended up being moved into the test directory, as the change didn't require it. We discussed it internally at some point, decided it was not necessary yet the move went in

Most of the test-machine doesn't need the admin-client. And in fact even the bits that do (e.g. creating test topics) can be done using the [rest-proxy these days](https://docs.confluent.io/5.5.0/kafka-rest/api.html#post--v3-clusters-(string-cluster_id)-topics)

PR #279, opened 2020-12-07
Forkify

cddr:
# Checklist

- [ ] tests


dakra:
Nice.
FYI, I have a personal fork [here](https://github.com/dakra/jackdaw) where I merged some PRs already.
Mainly because I wanted a newer Kafka Streams lib.
(You can also branch of this fork if you want)


Nice to see new development happening again :)

cddr:
Ah nice! I might just use yours then Daniel. I didn't actually mean to submit this as a PR into the upstream repo.

dakra:
Didn't mean you should use mine rather than just use it as a starting point for your fork.

I don't think I'm capable of maintaining a jackdaw fork. I think it's a good idea if you do a "real fork". At least until development here picks up again.

(Obviously you can use my fork if all you want is those few PR merged but I'm probably not doing any active development on it
and would rather use your fork ;)

PR #287, opened 2021-04-12
Fix docs

LouiseKlodt:
# Summary of changes
Fixes the code examples and documentation for `test-machine`

- [ ] tests

### DIFF
--- a/doc/test-machine.md
+++ b/doc/test-machine.md
@@ -32,7 +32,7 @@ to these services in a shared environment like uat/staging.
   (:require
     [my.app :as app]
     [jackdaw.serdes :refer [string-serde edn-serde]]
-    [jackdaw.serdes.json :as json-serde]
+    [jackdaw.serdes.json :as jsj]
     [jackdaw.test :refer [test-machine]]
     [jackdaw.test.transports :as trns]))
 
@@ -47,7 +47,7 @@ to these services in a shared environment like uat/staging.
 (def topic-config
   {:foo {:topic-name "foo"
          :key-serde (string-serde)
-         :value-serde (json-serde/serde)
+         :value-serde (jsj/serde)
          :partition-count 1
          :replication-factor 1}
    :bar {:topic-name "foo"
--- a/doc/test-machine.md
+++ b/doc/test-machine.md
@@ -40,6 +40,10 @@ to these services in a shared environment like uat/staging.
   {"bootstrap.servers" "localhost:9092"
    "group.id" "my-app"})
 
+(def remote-kafka-config
+  {:bootstrap-uri "my-real-rest-proxy-url"
+   :group-id "my-app"})
+
 (def topic-config
   {:foo {:topic-name "foo"
          :key-serde (string-serde)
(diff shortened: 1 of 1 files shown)


cddr:
Thanks for giving the docs some :heart: Louise. Particularly the note about `:partition-count`. That's a common stumbling block.

minimal (approved):

doc/test-machine.md
```diff
@@ -31,32 +31,42 @@ to these services in a shared environment like uat/staging.
 (ns my.app-test
   (:require
     [my.app :as app]
-    [jackdaw.test :as j.t]
+    [jackdaw.serdes :refer [string-serde edn-serde]]
+    [jackdaw.serdes.json :as json-serde]
```
gphilipp, line 35: I'm not a fan of the ns alias name, it's easily confused with a function name (like the 2 above). Can you name it `serdes.json` or even `jsj`

doc/test-machine.md
```diff
@@ -135,11 +146,43 @@ better that you write this macro yourself so that you can tailor it to your own
...
+  [f transport]
+  (with-fixtures [(topic-fixture kafka-config input-topic-config)
+                  (topic-fixture kafka-config output-topic-config)
+                  (service-ready? {:http-url "http://localhost:8082"
+                                   :timeout 5000})]
+    (with-open [machine (test-machine transport)]
       (f machine))))
 ```
+
+The `topic-fixture` function creates all the topics named in the supplied `topic-config` before running tests. You have to import `TopologyTestDriver` for `test-machine` to work, which requires you to bring in the `org.apache.kafka/kafka-streams-test-utils` library as a dependency, using a version within `2.0.0` - `2.3.0`. The `topic-fixture` expects the `topic-config` to contain `:partition-count` and `:replication-factor` to be present, besides the `topic-name` and key-value serdes.
```
gphilipp, line 188: Why not above than `2.3.0`?

gphilipp (changes requested):

doc/test-machine.md
LouiseKlodt: OK, makes sense, thanks!

doc/test-machine.md
LouiseKlodt: Thanks @gphilipp  - Currently, when bringing in versions of `kafka-streams-test-utils` higher than `2.3.0`,  it throws errors. For example, for `2.4.0`:
```
Syntax error (ClassNotFoundException) compiling new at (jackdaw/test.clj:250:5).
org.apache.kafka.streams.internals.QuietStreamsConfig
```
I assume we need to bring Jackdaw up to date with the newest versions from Apache Kafka's libraries.

I tried all version numbers from `2.4.0, 2.5.0` ... up to the latest `2.8.0` - they all throw errors in `jackdaw/test.clj`.

I tested this in a dummy example app, you can find it here: 
https://github.com/FundingCircle/jackdaw-examples

gphilipp: Ok then, let's file a separate issue for that then 👍 .

gphilipp (approved):

doc/test-machine.md
LouiseKlodt: OK, opened an issue here: https://github.com/FundingCircle/jackdaw/issues/290

Issue #295, opened 2021-06-14
Fix Jackdaw Client API documentation to use poll instead of log


LouiseKlodt:
The official documentation for the Client API (Consumer) currently advocates to use  the `jackdaw.client.log/log` function to retrieve messages. However, when working with the Consumer API I encountered several problems when using `jackdaw.client.log/log`. It looks like the `log` function is intended to be used for testing purposes only. It doesn't allow the consumer to commit offsets after each poll, for example, and the function itself is not tested in Jackdaw.

The correct function to use in production code, seems to be `poll`, in combination with a looping construct. 

Proposed changes:
Fix the documentation for the Client API (Consumer) to use the `poll` method, and add a working code example demonstrating its usage. Working on the fixes on branch `louiseklodt/fix-consumer-docs-for-poll`

PR #296, opened 2021-06-17
Fix Client API docs to use poll instead of log and add sample app for Client API

LouiseKlodt:
Fixes issue https://github.com/FundingCircle/jackdaw/issues/295

-> fix the Client API documentation to use the `poll` function instead of `log`
-> add working code example, to show how to construct the consumer polling loop. 
-> add `rolldice` sample app for Client API to `examples` folder


gphilipp (approved):
Thanks @LouiseKlodt, great doc update and good example too 👏

minimal (approved):
Looks great, thanks!

PR #305, opened 2021-10-22
Some additions and helpers for Kafka Streams interop

99-not-out:
* added simple helpers in `lambdas` for using transformers more easily in a stream
* added the `flatTransform` and `flatTransfromValues` calls to the core streams interface
* added a helper function to add KV state stores to a streams builder

### DIFF
--- a/src/jackdaw/streams/interop.clj
+++ b/src/jackdaw/streams/interop.clj
@@ -25,6 +25,8 @@
             ValueMapper ValueMapperWithKey ValueTransformerSupplier Windows]
            [org.apache.kafka.streams.processor
             StreamPartitioner]
+           [org.apache.kafka.streams.state
+            KeyValueStore Stores]
            (org.apache.kafka.streams.processor.api
             ProcessorSupplier)))
 
@@ -129,6 +131,15 @@
                    ^String topic-name
                    ^Consumed (topic->consumed topic-config))))
 
+  (with-kv-state-store
+    [builder {:keys [store-name key-serde value-serde] :as store-config}]
+    (.addStateStore ^StreamsBuilder streams-builder
+                    (Stores/keyValueStoreBuilder
+                     (Stores/persistentKeyValueStore store-name)
+                     key-serde
+                     value-serde))
+    builder)
+  
   (streams-builder*
     [_]
     streams-builder))
@@ -335,6 +346,17 @@
                  ^TransformerSupplier (transformer-supplier transformer-supplier-fn)
                  ^"[Ljava.lang.String;" (into-array String state-store-names))))
 
+  (flat-transform
+    [this transformer-supplier-fn]
+    (flat-transform this transformer-supplier-fn []))
+
+  (flat-transform
+    [_ transformer-supplier-fn state-store-names]
+    (clj-kstream
+     (.flatTransform ^KStream kstream
+                     ^TransformerSupplier (transformer-supplier transformer-supplier-fn)
+                     ^"[Ljava.lang.String;" (into-array String
+                                                        (clojure.core/map name state-store-names)))))
   (transform-values
     [this value-transformer-supplier-fn]
     (transform-values this value-transformer-supplier-fn []))
@@ -345,6 +367,18 @@
      (.transformValues ^KStream kstream
                        ^ValueTransformerSupplier (value-transformer-supplier value-transformer-supplier-fn)
                        ^"[Ljava.lang.String;" (into-array String state-store-names))))
+
+  (flat-transform-values
+    [this value-transformer-supplier-fn]
+    (flat-transform-values this value-transformer-supplier-fn []))
+
+  (flat-transform-values
+    [_ value-transformer-supplier-fn state-store-names]
+    (clj-kstream
+     (.flatTransformValues ^KStream kstream
+                           ^ValueTransformerSupplier (value-transformer-supplier value-transformer-supplier-fn)
+                           ^"[Ljava.lang.String;" (into-array String
+                                                              (clojure.core/map name state-store-names)))))
 
   (join-global
     [_ global-ktable key-value-mapper-fn joiner-fn]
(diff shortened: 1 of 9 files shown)


src/jackdaw/streams/extras.clj
```diff
@@ -3,13 +3,117 @@
...
+  (fn []
+    (let [ctx (atom nil)]
+      (reify ValueTransformer
+        (init [_ processor-context]
+          (reset! ctx processor-context))
+        (close [_])
+        (transform [_ v]
+          (xfm @ctx v))))))
+
+(defn with-state-store
```
minimal, line 68: something like `with-kv-state-store` or mentioning in the docstring would be more informative since it's not the only type of state store

99-not-out: 👍

src/jackdaw/streams/extras.clj
```diff
@@ -3,13 +3,117 @@
...
+           (org.apache.kafka.streams.kstream KStream KTable KGroupedStream KGroupedTable
+                                             GlobalKTable TimeWindowedKStream
+                                             SessionWindowedKStream Transformer ValueTransformer)
+           (org.apache.kafka.streams.processor ProcessorContext StateRestoreListener)
+           (org.apache.kafka.streams.state Stores)
            [java.time Instant Duration]))
 
 (set! *warn-on-reflection* true)
 
+(defn transformer-with-ctx
```
mattford63, line 22: Hi :-)

Do we care about consistency with how the processor function is dealt with?
- this is now not in the same namespace so perhaps awkward to find
- processor natively sorts the context out for you but transform doesn't - is backwards compatibility a consideration here?  Can we introduce a breaking change for consistent behaviour?  Or call it `transfrom!` maybe?
- the implementation with reify in this change is different to the implementation for processor which we based our changes on see `FnTransformer` and `transformer` here https://github.com/FundingCircle/originations-application-tracker/blob/application-creation-command-handler/src/tracker/stream_utils.clj which was based on https://github.com/FundingCircle/jackdaw/blob/30b341d74c42dc0049221648113ac43acd572115/src/jackdaw/streams/lambdas.clj#L145

99-not-out: did some moving around which resolves these (very valid) points hopefully

kidpollo:
I might work with this one as this one looks newer :P

99-not-out:
> I might work with this one as this one looks newer :P

Yes @kidpollo  - this one will be the merged version of these additions (and the other older PR is closed out now)

test/jackdaw/streams_test.clj
```diff
@@ -475,6 +470,41 @@
...
+                                          (k/flat-transform transformer-supplier-fn)
+                                          (k/to topic-b))))
+          publish (partial mock/publish driver topic-a)]
+
+      (publish 1 1)
+      (publish 1 2)
+      (publish 1 4)
+
+      (let [keyvals (mock/get-keyvals driver topic-b)]
+        (is (= 6 (count keyvals)))
```
gphilipp, line 500: That's quite verbose, why not asserting the whole value of the data structure at once?

test/jackdaw/streams_test.clj
```diff
@@ -1335,3 +1399,61 @@
...
+        (let [publisher (partial mock/publish driver input-t)]
+          
+          (publisher 100 {:val 10})
+
+          (let [[[k v]] (mock/get-keyvals driver output-t)]
+            (is (= 11 (:new-val v)))
+            (is (= "input-topic" (:topic v)))))))))
+
+(deftest with-kv-state-store-test
+  (testing "Transfromer with state store sugar"
```
gphilipp, line 1424: Typo 💅

src/jackdaw/streams/interop.clj
```diff
@@ -129,6 +131,15 @@
                    ^String topic-name
                    ^Consumed (topic->consumed topic-config))))
 
+  (with-kv-state-store
+    [builder {:keys [store-name key-serde value-serde] :as store-config}]
+    (.addStateStore ^StreamsBuilder streams-builder
+                    (Stores/keyValueStoreBuilder
+                     (Stores/persistentKeyValueStore store-name)
```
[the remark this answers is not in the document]
99-not-out: yeah - there are several variations for KV stores, and a follow up can grab a few of them at once

kidpollo:
Cool cool thanks:)

test/jackdaw/streams_test.clj
99-not-out: Personally I find it easier to debug when one part of the seq goes wrong (you're not trying to work out yourself which part is incorrect). Although granted in this example it's likely to more spectacularly break in a regression.

src/jackdaw/streams/mock.clj
```diff
@@ -38,29 +29,6 @@
...
-        (configure ::test-driver-time (inc time)))))
-
-(defn collect
-  "Collects the test results. The test driver returns a list of messages with
-  each message formatted like \"key:value\""
-  [streams-builder]
-  (let [processor-supplier (-> streams-builder config ::processor-supplier)
-        processed (vec (.processed processor-supplier))]
-    (.clear (.processed processor-supplier))
-    processed))
```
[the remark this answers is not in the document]
gphilipp: We decided it's not needed to deprecate in the `mock` ns.

src/jackdaw/streams/mock.clj
```diff
@@ -17,19 +16,11 @@
...
 (set! *warn-on-reflection* false)
 
-(defn streams-builder
-  "Creates a mock streams-builder."
-  ([]
-   (streams-builder (interop/streams-builder)))
-  ([streams-builder]
-   (configured/streams-builder
-    {::streams-builder (k/streams-builder* streams-builder)}
-    streams-builder)))
```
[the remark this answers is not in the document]
gphilipp: We decided it's not needed to deprecate in the mock ns.

gphilipp (approved):
LGTM

gphilipp:
Let's add a CHANGELOG entry

gphilipp (approved):

minimal (approved):

PR #316, opened 2022-05-20
[TDI-1361] make the avro serialization exception message more verbose

apanourgiasfc:
Avro error messages now include the relevant path and topic.

On type errors:
- the path points to the attribute that caused the error
- we'll also include the data of said attribute

On any other error (e.g missing/unknown):
- the path points to the map the attribute that caused the error belongs

**Error message examples**

Incorrect Union type - `:my-attribute` is an enum and we supply a string:
```
{:message "java.lang.String is not a valid type for union [NULL, ENUM]", :path [:my-attribute], :data "Something", :topic "topic-name-1"}
```

Incorrect type for nested field `:name` of map `:person`:
```
{:message "java.lang.Long is not a valid type for string", :topic "topic-name-1", :path [:person :name], :data 123}

```

Unknown field `:application-id2` in top level:
```
{:message "Field application_id2 not known in topic_name", :topic "topic-name-1", :path []}
```

Unknown nested field `:name2` in `:person` map:
```
{:message "Field name2 not known in topic_name", :topic "topic-name-1", :path [:person], :data {:name "John", :name2 "dasds", :surname "Doe"}, :path [:person]}
```

Missing required field `:message-id` in top level:
```
{:message "Field message_id type:STRING pos:0 not set and has no default value", :topic "topic-name-1", :path []}
```

Missing nested required field `:name` in `:person` map:
```
{:message "Field name type:STRING pos:0 not set and has no default value", :topic "topic-name-1", :path [:person]}

```

### DIFF
--- a/src/jackdaw/serdes/avro.clj
+++ b/src/jackdaw/serdes/avro.clj
@@ -71,7 +71,7 @@
            [java.io ByteArrayOutputStream ByteArrayInputStream]
            [java.util Collection Map UUID]
            [org.apache.avro
-            Schema$Parser Schema$ArraySchema Schema Schema$Field]
+            AvroTypeException Schema$Parser Schema$ArraySchema Schema Schema$Field]
            [org.apache.avro.io
             EncoderFactory DecoderFactory JsonEncoder]
            [org.apache.avro.generic
@@ -149,7 +149,8 @@
 (defn validate-clj! [this x path expected-type]
   (when-not (match-clj? this x)
     (throw (ex-info (serialization-error-msg x expected-type)
-                    {:path path, :data x}))))
+                    {:path path, :data x}
+                    (AvroTypeException. "Type Error")))))
 
 ;;;; Primitive Types
 
@@ -428,7 +429,8 @@
   (clj->avro [_ clj-map path]
     (when-not (map? clj-map)
       (throw (ex-info (serialization-error-msg clj-map "record")
-                      {:path path, :clj-data clj-map})))
+                      {:path path, :clj-data clj-map}
+                      (AvroTypeException. "Type Error"))))
 
     (let [record-builder (GenericRecordBuilder. schema)]
       (try
@@ -485,7 +487,8 @@
                                                     (map #(.getType ^Schema %))
                                                     (str/join ", ")
                                                     (format "union [%s]")))
-                      {:path path, :clj-data clj-data})))))
+                      {:path path, :clj-data clj-data}
+                      (AvroTypeException. "Type Error"))))))
 
 (defn ->UnionType
   "Wrapper by which to construct a `UnionType` which handles the
@@ -523,10 +526,16 @@
                               (try
                                 (.serialize base-serializer topic (clj->avro coercion-type data []))
                                 (catch clojure.lang.ExceptionInfo e
-                                  (let [data (-> e
-                                                 ex-data
-                                                 (assoc :topic topic :clj-data data))]
-                                    (throw (ex-info (.getMessage e) data))))))}
+                                  (let [exception-data    (-> e
+                                                              ex-data
+                                                              (assoc :topic topic :clj-data data))
+                                        path              (:path exception-data)
+                                        exception-message (str (merge {:message (.getMessage e)
+                                                                       :topic   (:topic exception-data)
+                                                                       :path    path}
+                                                                      (when (instance? org.apache.avro.AvroTypeException (ex-cause e))
+                                                                        {:data (get-in data path)})))]
+                                    (throw (ex-info exception-message exception-data))))))}
         clj-serializer (fn/new-serializer methods)]
     (.configure ^Serializer clj-serializer serializer-config key?)
     clj-serializer))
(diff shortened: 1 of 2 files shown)


src/jackdaw/serdes/avro.clj
```diff
@@ -525,8 +525,14 @@
                                 (catch clojure.lang.ExceptionInfo e
                                   (let [data (-> e
                                                  ex-data
-                                                 (assoc :topic topic :clj-data data))]
-                                    (throw (ex-info (.getMessage e) data))))))}
+                                                 (assoc :topic topic :clj-data data))
+                                        path (:path data)
```
gphilipp, line 529: remove this let binding and then do 👇

src/jackdaw/serdes/avro.clj
```diff
@@ -525,8 +525,14 @@
...
+                                        info (str (merge {:message (.getMessage e)
+                                                          :topic   (:topic data)}
+                                                         (when (not (empty? path))
```
gphilipp, line 532:
```clojure
(when-let [path (:path data)]
   {:path path
    :data (get-in (:clj-data data) path)})
```
better cohesion 💁‍♂️

src/jackdaw/serdes/avro.clj
```diff
@@ -525,8 +525,14 @@
                                 (catch clojure.lang.ExceptionInfo e
                                   (let [data (-> e
                                                  ex-data
-                                                 (assoc :topic topic :clj-data data))]
-                                    (throw (ex-info (.getMessage e) data))))))}
+                                                 (assoc :topic topic :clj-data data))
```
gphilipp, line 528: Appreciate it was already in the previous version of the code, but can we rename either the left or right hand-side `data` to remove the variable shadowing?

gphilipp (changes requested):

src/jackdaw/serdes/avro.clj
apanourgiasfc: Sure can move it into a when-let with a seq. The path is never nil, if not used it's an empty vector.

src/jackdaw/serdes/avro.clj
apanourgiasfc: 👍

src/jackdaw/serdes/avro.clj
gphilipp: Yeah good catch, so 
```clojure
when-let [path (-> data :path seq)]
    {:path path
     :data (get-in (:clj-data data) path)})
```

apanourgiasfc: I made any changes and expanded the PR description. The one thing that I don't like is that for errors such as missing required field or unknown field, the path is pointing to the map said fields are missing from or are found. Which is different to the behaviour for the type errors. So if something is nested, with this code you might end up getting the data for the whole map that contains any attribute with erroneous data. We could just omit the data and always include the path in the message which would be more consistent, or just always include the data under any path (which would make messages very verbose - as most errors are on the top level usually)

99-not-out:
LGTM

99-not-out (approved):

src/jackdaw/serdes/avro.clj
```diff
@@ -523,16 +523,15 @@
...
-                                                            :data (get-in (:clj-data data) path)})))]
-                                    (throw (ex-info info data))))))}
+                                  (let [exception-data (-> e
+                                                           ex-data
+                                                           (assoc :topic topic :clj-data data))
+                                        exception-message (str (merge {:message (.getMessage e)
+                                                                       :topic   (:topic exception-data)}
+                                                                      (when-let [path (seq (:path exception-data))]
+                                                                        {:path (vec path)
+                                                                         :data (get-in (:clj-data exception-data) path)})))]
```
gphilipp, line 533: I think you can replace `(:clj-data exception-data)` with `data`

src/jackdaw/serdes/avro.clj
apanourgiasfc: I went for a slightly different approach. I always include the path in the error message and then the data are only included if it's a type error that was raised (as in any other case we don't really care what the data is)

src/jackdaw/serdes/avro.clj
```diff
@@ -149,7 +149,8 @@
 (defn validate-clj! [this x path expected-type]
   (when-not (match-clj? this x)
     (throw (ex-info (serialization-error-msg x expected-type)
-                    {:path path, :data x}))))
+                    {:path path, :data x}
```
gphilipp, line 152: shouldn't we use `:clj-data` instead of `data` for consistency with the other raised exceptions?

apanourgiasfc: sure, not changed by this pr but no harm in aligning

gphilipp: Just realised it's a breaking change 🤔 What do others think about it?

apanourgiasfc: Yeah it's changing the behaviour for sure.

gphilipp (approved):

gphilipp (approved):

PR #321, opened 2022-05-29
Port to kafka v3.2.0 and update deps

ghaskins:
Signed-off-by: Greg Haskins <greg@manetu.com>

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -7,38 +7,38 @@
 
   :repositories [["confluent" {:url "https://packages.confluent.io/maven/"}]]
 
-  :dependencies [[aleph "0.4.6"]
+  :dependencies [[aleph "0.4.7"]
                  [danlentz/clj-uuid "0.1.9"
                   :exclusions [primitive-math]]
 
                  ;; Confluent does paired releases with Kafka, this should tie
                  ;; off with the kafka version.
                  ;; See https://docs.confluent.io/current/release-notes.html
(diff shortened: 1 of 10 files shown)


gphilipp:
Hi @ghaskins, thanks for the PR this is very much appreciated. We've discussed this internally and would prefer to see it split into 3 PRs:

- one that allows to stick with 2.x but removes all calls to deprecated stuff and upgrades libs
- one that upgrades to 3.2.0
- last one to upgrade to clojure 1.11.1
 
I'll cherry pick changes from your branch if you don't mind (and give credit where it's due of course).

ghaskins:
@gphilipp the plan makes sense, and of course I don't mind.  Ty!

BTW: I actually tried a variation of your proposal (doing the 2.x with updates as a different patch) and it was inexplicably failing UTs.  I didn't investigate too deeply, though, since I had just finished getting the 3.2 port to work.  I did _not_ try to split clojure out, though.

mt3593:
Hi all, any update on this? Happy to help out if I can.

kidpollo:
@marcoccchan @99-not-out  this is also important. Specially for ppl with M1 macs. There is even newer kafka libs now. I also don't see a point of splitting this into multiple prs

minimal:
Thanks this was included in #353 and released in https://github.com/FundingCircle/jackdaw/releases/tag/0.9.10

PR #337, opened 2022-08-24
[#336] use a confluent version that properly supports kafka 2.8.0

### LINKED ISSUE #336: confluent and kafka version mismatch
as currently written there's an explicit mismatch between confluent and apache kafka versions:
- confluent is pinned to [version `6.1.6` ](https://github.com/FundingCircle/jackdaw/blob/master/project.clj#L18)which only supports [`kafka 2.7.0`](https://docs.confluent.io/platform/6.1.6/release-notes/index.html)
- kafka is pinned to [version `2.8.0`](https://github.com/FundingCircle/jackdaw/blob/master/project.clj#L22)

this isn't showing up as an issue because of the way leiningen resolves dependency conflicts. however, it does show up when trying to use this library as a gitlib via `tools.deps`.

`tools.deps` defaults to using the most recent version problems arise which causes an issue starting streams up since they now rely on the `org.apache.kafka.common.Uuid` class which doesn't exist before `2.8.0`.

darin-reify:
fixes #336

# Checklist

- [ ] tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Changelog
 
 ### Unreleased
+- use version of confluent platform that supports Kafka 2.8.0
 
 ### [0.9.6] - [2022-08-01]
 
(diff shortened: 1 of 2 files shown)


darin-reify:
closing for now as the tests don't pass due to some `jackson` issues and i don't have time to look into them.

PR #360, opened 2023-04-28
add manifold and keep aleph in dev dependencies

AndreaCrotti:
# Checklist

- [x] tests


AndreaCrotti:
I think maybe the only problem would be that people using the test-machine would just need to add the aleph dependency in their own project, but I think if that's just mentioned in the README it should not be an issue.

minimal (approved):

PR #367, opened 2024-01-15
Fix avro serialization of union types

DaveWM:
This is a fix for a bug in the avro serde, where serialization is sometimes ignoring default values that appear within a union type. This is only the case when the default value is `false` or `nil`.

The problem is that when `match-clj?` is called for a record type (as it is within `UnionType`'s `clj->avro` method), it checks whether any fields missing from the clojure data have a default value in the avro schema. To do this it calls `.defaultVal`, but if the default value is `nil` or `false` this returns falsey. I've therefore replaced this call with `.hasDefaultValue`. 

# Checklist

- [x] tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ### Unreleased
 
+- Fixed a bug with avro serialization, where default values within a union type would not be used correctly.
+
 ### [0.9.12] - [2023-12-05]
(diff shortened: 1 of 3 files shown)


fc-infosec-sast:
<h2>✅ SAST Report</h2>

No issues found. Good job! 💪

PR #376, opened 2024-08-12
Fix seek-to-timestamp

### LINKED ISSUE #371: Seek to timestamp possibly off by one
I'm still getting used to working with kafka, so I may have misunderstood the semantics, but it looks like seek-to-timestamp is handling the missing timestamp-offset case incorrectly: I don't think it should increment the end offset (see [this line of client.clj](https://github.com/FundingCircle/jackdaw/blob/3dc4b18e21c961ca8c12bfe5e8cf11066fbfa4a0/src/jackdaw/client.clj#L296)).

FundingCircle/jackdaw#198 mentions being concerned about duplicate messages, but if I've understood correctly the endOffsets on a set of partitions are already past the most recent comitted message- the kafka javadocs state they correspond to the offset of the last committed message plus one (i.e. the offset that will be assigned the next message).

Similarly, I'm pretty sure that [the test associated with this case](https://github.com/FundingCircle/jackdaw/blob/3dc4b18e21c961ca8c12bfe5e8cf11066fbfa4a0/test/jackdaw/client_test.clj#L364) should be looking for 10, not 11, since the final message of the test data has offset 9.

Could someone with more experience with kafka's semantics take a look?

marcoccchan:
This fixes #371 — thanks to @djeis97 for spotting.

### DIFF
--- a/src/jackdaw/client.clj
+++ b/src/jackdaw/client.clj
@@ -293,7 +293,7 @@
              {:keys [^OffsetAndTimestamp ts-offset end-offset]}] offsets]
       (let [offset (or (when ts-offset
                          (.offset ts-offset))
-                       (inc end-offset))]
+                       end-offset)]
         (log/infof "Setting starting offset (topic=%s, partition=%s): %s"
                    (.topic topic-partition)
                    (.partition topic-partition)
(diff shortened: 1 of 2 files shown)


fc-infosec-sast:
<h2>✅ Coberos SAST Report</h2>

No issues found. Good job! 💪

minimal (approved):

PR #377, opened 2024-08-16
Retrieve end offsets before searching for offsets for timestamp

marcoccchan:
- this fixes a race condition where a message could be produced after searching for offsets of a timestamp but before end offsets are retrieved.


fc-infosec-sast:
<h2>✅ Coberos SAST Report</h2>

No issues found. Good job! 💪

minimal (approved):

Issue #378, opened 2024-08-23
kafka-streams-test-utils dependency missing


retzkek:
`org.apache.kafka/kafka-streams-test-utils` is getting included with scope `test`, so isn't included in the distribution, but is needed by jackdaw.data since 0.9.5 (https://github.com/FundingCircle/jackdaw/blob/master/src/jackdaw/data/producer.clj#L10). 

https://repo.clojars.org/fundingcircle/jackdaw/0.9.12/jackdaw-0.9.12.pom:
```xml
<dependency>
<groupId>org.apache.kafka</groupId>
<artifactId>kafka-streams-test-utils</artifactId>
<version>3.3.2</version>
<scope>test</scope>
</dependency>
```

Observed behavior:
```
$ lein repl                          
#error {                                                            
 :cause org.apache.kafka.streams.test.TestRecord    
 :via                                                               
 [{:type clojure.lang.Compiler$CompilerException
   :message Syntax error macroexpanding at (jackdaw/data/producer.clj:7:1).
   :data #:clojure.error{:phase :execution, :line 7, :column 1, :source jackdaw/data/producer.clj}
   :at [clojure.lang.Compiler load Compiler.java 7665]}
  {:type java.lang.ClassNotFoundException
   :message org.apache.kafka.streams.test.TestRecord
   :at [java.net.URLClassLoader findClass URLClassLoader.java 445]}] 
...
jackdaw-test.core=> (require 'jackdaw.admin)
Syntax error compiling at (jackdaw/admin.clj:1:1).
namespace 'jackdaw.data' not found
```
with only the following dependencies:
```
  :dependencies [[org.clojure/clojure "1.11.1"]
                 [fundingcircle/jackdaw "0.9.12"]]
```

But add `[org.apache.kafka/kafka-streams-test-utils "3.3.2"]` to the project and it works as expected.

I don't know enough about java/maven packaging to suggest the correct fix.

retzkek:
Oh, maybe it's as simple as removing it from the `:dev` dependencies, that must be overriding the scope (it's also in the default project dependencies)? https://github.com/FundingCircle/jackdaw/blob/master/project.clj#L80

```
$ git diff
diff --git a/project.clj b/project.clj
index 550b5b4..9d0862f 100644
--- a/project.clj
+++ b/project.clj
@@ -77,7 +77,6 @@
               :injections [(require 'io.aviso.logging.setup)]
               :dependencies [[io.aviso/logging "1.0"]
[...]

PR #384, opened 2026-09-01
Bump dependencies

alex-mckenna:
A meaningful refresh of dependencies / tooling around this repo. Main points are moving from `lein` to `clojure-cli` and bumping Kafka from 3.x to 4.x.

# Checklist

- [x] tests


build.clj
```diff
@@ -0,0 +1,100 @@
...
+
+          :else "0.0.0-SNAPSHOT"))))
+
+(def version (derive-version))
+(def class-dir "target/classes")
+(def jar-file (format "target/%s-%s.jar" (name lib) version))
+(def snapshot? (str/ends-with? version "-SNAPSHOT"))
+
+(defn- basis []
+  (b/create-basis {:project "deps.edn"}))
```
minimal:
```suggestion
  (b/create-basis {:root nil :project "deps.edn"}))
```
Lein build had `:scope "provided"` but this currently pins to the build container's pins. `:root nil` makes it `<scope>provided</scope>.` in the pom

build.clj
```diff
@@ -0,0 +1,100 @@
...
+  Credentials are read from the CLOJARS_USERNAME / CLOJARS_PASSWORD environment
+  variables. Non-snapshot releases are GPG signed."
+  [_]
+  (when-not (.exists (io/file jar-file))
+    (jar nil))
+  (dd/deploy {:installer :remote
+              :artifact jar-file
+              :pom-file (b/pom-path {:class-dir class-dir :lib lib})
+              :sign-releases? (not snapshot?)})
+  (println "Deployed" jar-file "to Clojars"))
```
minimal, line 117:
```suggestion
(defn deploy [_]
  (when-not (.exists (io/file jar-file))
    (jar nil))
  ;; deps-deploy 0.2.5 calls gpg/read-passphrase unconditionally in sign!, even
  ;; when a key id is supplied. System/console is nil on a non-TTY, so this NPEs
  ;; in CI. bin/gpg supplies the passphrase via loopback pinentry, so the value
  ;; read here is never used.
  (with-redefs [deps-deploy.gpg/read-passphrase (constantly "")]
    (dd/deploy {:installer :remote
                :artifact jar-file
                :pom-file (b/pom-path {:class-dir class-dir :lib lib})
                :sign-key-id "fundingcirclebot@fundingcircle.com"
                :sign-releases? (not snapshot?)}))
  (println "Deployed" jar-file "to Clojars"))
```
Without this it looks like you get `NullPointerException: Cannot invoke "java.io.Console.readPassword(String, Object[])" because "console" is null`. 
Also needs `[deps-deploy.gpg]` in the ns require

src/jackdaw/client.clj
```diff
@@ -211,6 +211,17 @@
   (doto ^Consumer consumer
     (.seek ^TopicPartition (jd/as-TopicPartition topic-partition) offset)))
 
+(defn- poll-for-assignment
```
minimal: This counts via iterations rather than wall clock time so 100 iterations can hit the limit very quickly during a rebalance with fast responses. It also fails silently. Can you port the one from the 4.2 branch that uses wall clock and throws a timeout ex-info https://github.com/FundingCircle/jackdaw/commit/1b5a148
Also it's duped in kafka.clj

CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Changelog
 
 ### Unreleased
+- **Kafka 4.x / Confluent Platform 8.x support.** Upgraded `org.apache.kafka` to 4.3.0 and the Confluent serializers/schema-registry client to 8.3.1. Adapts to the removal of `KStream.branch`/`through`/`transform*` (now built on `split`/`process`/`processValues`), `StreamPartitioner.partition` -> `partitions`, `AdminClient.alterConfigs` -> `incrementalAlterConfigs`, `Duration`-only `Consumer.poll`, the new `StreamsUncaughtExceptionHandler`, and the relocated `org.apache.kafka.tools.StreamsResetter`. The Docker test stack now runs Kafka 4.x in KRaft mode (no ZooKeeper). [#384](https://github.com/FundingCircle/jackdaw/pull/384)
+- **Migrated build tooling from Leiningen to `deps.edn`/`tools.build`.** Replaced `project.clj` with `deps.edn` + `build.clj`; example projects and CI updated accordingly. [#384](https://github.com/FundingCircle/jackdaw/pull/384)
```
minimal, line 5: Can add that this closes #378

deps.edn
```diff
@@ -0,0 +1,69 @@
...
+                org.apache.kafka/kafka-clients]},
+  io.confluent/kafka-avro-serializer
+  {:mvn/version "8.3.1",
+   :exclusions [org.apache.kafka/kafka-clients]},
+  io.confluent/kafka-json-schema-serializer
+  {:mvn/version "8.3.1",
+   :exclusions [org.apache.kafka/kafka-clients]},
+  org.apache.kafka/kafka-clients {:mvn/version "4.3.0"},
+  org.apache.kafka/kafka-streams {:mvn/version "4.3.0"},
+  org.apache.kafka/kafka-streams-test-utils {:mvn/version "4.3.0"},
```
jeremyvdw: Why not bumping to [4.3.1](https://kafka.apache.org/community/downloads/#431) right away ?

build.clj
```diff
@@ -28,17 +28,24 @@
...
-                            "--match" "[0-9]*.[0-9]*.[0-9]*")]
+                            "--match" "[0-9]*.[0-9]*.[0-9]*")
+            snapshot-tag (git "describe" "--tags" "--exact-match"
+                              "--match" "publish-snapshot-[0-9]*.[0-9]*.[0-9]*")]
         (cond
           exact-tag exact-tag
 
+          snapshot-tag
+          (let [[_ semver] (re-find #"publish-snapshot-(\d+\.\d+\.\d+)" snapshot-tag)]
+            (format "%s-SNAPSHOT" semver))
```
minimal, line 47: question: this changes the meaning of existing tag names. The old lein :status-to-version incremented the patch, so publish-snapshot-0.9.14 published 0.9.15-HEAD-SNAPSHOT; here the same tag would publish 0.9.14-SNAPSHOT. Is this intentional?

alex-mckenna: Yeah, I found it a bit counter-intuitive that snapshots would always be given a patch version instead of always matching the intended next version. I think it makes more sense to commit with the desired next version, then publish a snapshot from the branch so the numbers match.

alex-mckenna: @minimal If you have any suggestions for guards around this, I'm all ears

minimal: yeah I think it  makes sense. we can always tweak it later if needed

minimal (approved):
nice work

